Issue #23, opened 2015-04-21
host routing


jeffnappi:
Would be nice to have the ability to use hostnames to route requests. This would allow the use of configurable-http-proxy to route to a demo application root.

PR #24, opened 2015-04-22
host routing support

jeffnappi:
Adds host routing per #23.

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -88,6 +88,14 @@ var authorized = function (method) {
     };
 };
 
+var parse_host = function(req) {
+    var host = req.headers['host'];
+    if (host) {
+        host = host.split(':')[0];
+    }
+    return host;
+};
+
 var ConfigurableProxy = function (options) {
     var that = this;
     this.options = options || {};
@@ -95,6 +103,8 @@ var ConfigurableProxy = function (options) {
     this.auth_token = this.options.auth_token;
     this.includePrefix = options.includePrefix === undefined ? true : options.includePrefix;
     this.routes = {};
+    this.host_routing = this.options.host_routing;
+
     if (this.options.default_target) {
         this.add_route('/', {
             target: this.options.default_target
@@ -148,6 +158,9 @@ var ConfigurableProxy = function (options) {
 ConfigurableProxy.prototype.add_route = function (path, data) {
     // add a route to the routing table
     path = trie.trim_prefix(path);
+    if (this.host_routing && path != '/') {
+        data['host'] = path.split('/')[1];
+    }
     this.routes[path] = data;
     this.trie.add(path, data);
     this.update_last_activity(path);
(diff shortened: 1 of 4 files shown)


minrk:
Can you explain the use case for this?

jeffnappi:
Hi @minrk.

This can be used along with a modified tmpnb to serve up temporary containers utilizing a wildcard subdomain rather than a path, allowing one to use unmodified applications that require that they live at the URL root.

e.g. vilfsypfs9ey.demo.application.org -> http://localhost:8409

rgbkrk:
Would this mean that certs would have to be a wildcard across a domain (or SNI across multiples)?

minrk:
@jeffnappi ah, I get it. I had misread where the host info was coming from. Thanks for clarifying.

@rgbkrk presumably one would have an appropriate wildcard cert, or do SSL termination prior to the proxy.

jeffnappi:
@minrk no worries :)

@rgbkrk if the SSL termination is happening at the proxy, it would need to have an appropriate wildcard SSL cert, presumably w/ appropriate "Subject Alternate Names" specified. e.g. could be *.domain.com with SAN of *.demo.domain.com.

minrk:
@jeffnappi I fixed the dependency problem. Can you rebase on master so we can see the tests run again?

jeffnappi:
@minrk I rebased and pushed, still seems to fail on Travis CI.

Here's the test output for me:

```
$ jasmine-node .
.....................

Finished in 0.787 seconds
21 tests, 122 assertions, 0 failures, 0 skipped
```

minrk:
Looks like the test failures are real regressions from this PR - extra host keys not expected.

minrk:
When I run the tests locally with `npm test`, I see the same failures as on Travis.

jeffnappi:
@minrk I see the issue, I will update and push :)

lib/configproxy.js
```diff
@@ -221,11 +234,13 @@ ConfigurableProxy.prototype.delete_routes = function (req, res, path) {
...
 
-ConfigurableProxy.prototype.target_for_url = function (url) {
+ConfigurableProxy.prototype.target_for_req = function (req) {
     // return proxy target for a given url path
-    var route = this.trie.get(url);
+    var base_path = (this.host_routing) ? '/' + parse_host(req) : '';
+    var route = this.trie.get(base_path + req.url);
     if (route) {
         return {
+            host: route.host,
```
minrk: is there a reason to return host here? It doesn't seem to be used anywhere.

test/api_spec.js
```diff
@@ -33,26 +33,32 @@ describe("API Tests", function () {
     it("Basic proxy constructor", function () {
         expect(proxy).toBeDefined();
         expect(proxy.default_target).toBe(undefined);
-        expect(proxy.target_for_url('/')).toEqual({
+        expect(proxy.target_for_req({url: '/'})).toEqual(
+          jasmine.objectContaining({
```
minrk: I'd rather keep exact-match tests here, rather than containing tests, if we can. If these are the only keys we should have, but host is added as undefined, then the tests was rightly failing and host probably shouldn't be added, rather than adding host and making the tests insensitive to change.

jeffnappi:
@minrk good call, I will clean that up and update the PR.

rgbkrk:
Looking good, thanks for coming back to this rapidly.

dgonzo:
@minrk Could this work for non-port 80 requests if the port number were added to the hostRouting function?

https://github.com/jupyterhub/configurable-http-proxy/blob/53310e4ca7465d2dfacfc8265d4a6f075bbeef3e/test/proxy_spec.js#L190-L203

We often serve demo apps on ports other than 80 or 443.

minrk:
@dgonzo can you clarify a bit what you are after? Host routing should work regardless of the port CHP is listening on.

PR #28, opened 2015-04-28
handle encoded URIs

minrk:
- routing table contains only decoded URIs
- URIs are decoded before checking with trie


Carreau:
+1, when do you start using typescript ?

minrk:
@Carreau ha! I'll probably rewrite it in Go before I do that. Might be worth an experiment, though.

rgbkrk:
> I'll probably rewrite it in Go before I do that.

Hahahahaha.

I'm excited by the prospect of doing this, since you already have an API spec to adhere to.

PR #29, opened 2015-04-28
Include iojs in the test suite

rgbkrk:
Ensure we're working properly with iojs.


minrk:
Do we need more than one iojs?

rgbkrk:
One's a tagged version, the other is the current stable release. Up to you.

minrk:
I care little enough about it that I think just current stable is enough.

minrk:
Since iojs takes 10x as long as nodejs to set up on travis, I don't think we need two of them. Can you leave it as just `iojs`?

rgbkrk:
Ew. Sure.

Issue #30, opened 2015-05-06
Upgrading to HTTPS automatically.


rgbkrk:
For the sake of JupyterHub, it would be really convenient if on simple deployments using certs if there was a way to forward port 80 to 443. Perhaps this could be an extra flag that runs a simple redirect server alongside the other services?

```node
var http = require('http');
http.createServer(function (req, res) {
    res.writeHead(301, { "Location": "https://" + req.headers['host'] + req.url });
    res.end();
}).listen(80);
```

Not sure how best to expose this as an option or if you would want this in here.

minrk:
Great idea. It's always bugged me that we can't do this with tornado.

rgbkrk:
Ok, cool. I'll draft a PR.

PR #31, opened 2015-05-06
Flag and server for upgrading from 80 to https on host

rgbkrk:
Addresses #30.

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -25,7 +25,7 @@ args
     .option('--api-ssl-key <keyfile>', 'SSL key to use, if any, for API requests')
     .option('--api-ssl-cert <certfile>', 'SSL certificate to use, if any, for API requests')
     .option('--default-target <host>', 'Default proxy target (proto://host[:port]')
-    .option('--redirect-to-https-from-port <redirect-port>', 'Redirects traffic from redirect port to the port')
+    .option('--redirect-port <redirect-port>', 'Redirect HTTP requests on this port to the server on HTTPS')
     // passthrough http-proxy options
     .option('--no-x-forward', "Don't add 'X-forward-' headers to proxied requests")
     .option('--no-prepend-path', "Avoid prepending target paths to proxied requests")
@@ -76,7 +76,7 @@ if (args.apiSslKey || args.apiSslCert) {
 options.default_target = args.defaultTarget;
 options.host_routing = args.hostRouting;
 options.auth_token = process.env.CONFIGPROXY_AUTH_TOKEN;
-options.redirectPort = args.redirectToHttpsFromPort;
+options.redirectPort = args.redirectPort;
 
 // certs need to be provided for https redirection
 if (!options.ssl && options.redirectPort) {
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -25,7 +25,7 @@ args
     .option('--api-ssl-key <keyfile>', 'SSL key to use, if any, for API requests')
     .option('--api-ssl-cert <certfile>', 'SSL certificate to use, if any, for API requests')
     .option('--default-target <host>', 'Default proxy target (proto://host[:port]')
-    .option('--upgrade-to-https', 'Upgrade to HTTPS from port 80 -> port')
+    .option('--redirect-to-https-from-port <redirect-port>', 'Redirects traffic from redirect port to the port')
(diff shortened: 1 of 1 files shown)


minrk:
It seems like there should be some configurability of ports here. I see two common cases:

1. standard ports (http on 80, https on 443)
2. http requests to a custom port (e.g. 8000)

Do you see a way to address 2 as well?

rgbkrk:
> It seems like there should be some configurability of ports here.

I wondered about this. Potential options:

```
--redirect-to-https-from-port <port>

--redirect-to-port <port>
```

The to port would implicitly be the port the main config proxy runs on `--port`

There might also be the case where people run the server on e.g. 8443 and they need the connection upgraded (not redirected). This falls under the same awkwardness we face with a locally running tornado server running on 8888 but serving https. `req.headers['host']` would contain the port if it came from. I think that's a separate issue.

rgbkrk:
When the server is reached via 80, certs are configured, and is listening on 443 we should do the redirect automatically. If it's on a non standard port we should do an upgrade which is a separate issue entirely.

After working on logic for handling redirected ports, it doesn't seem worth it to maintain code that will redirect 8000 -> 8443 (as an example). I would have made the command line argument shorter (redirect port) but I didn't want to convey it as a direct http port redirection (sans https).

rgbkrk:
To do the upgrade to https on a random high port (as opposed to the browser being smart about a redirect from 80 -> 443 by convention), we'd have to have a TCP proxy at the front that can notice a TLS/SSL handshake and choose whether to hit the redirection "server" or the main https-backed proxy.

Related [answer on SO](http://stackoverflow.com/a/23975955/700228).

minrk:
Curious: what happens if I make an https request to the redirect port?

rgbkrk:
> Curious: what happens if I make an https request to the redirect port?

I'll assume you mean as part of this PR.

If you request it with `curl`, it will report back that the server aborted the SSL handshake. However, if you do it with Chrome it figures it out and tries to go over HTTP (which then leads to the redirect and the right endpoint).

minrk:
Okay, I think that makes sense.

bin/configurable-http-proxy
```diff
@@ -107,3 +109,17 @@ log.info("Proxy API at %s://%s:%s/api/routes",
...
     listen.api_port);
 
+if (options.redirectPort) {
+  var http = require('http');
+  
+  http.createServer(function (req, res) {
+    var host = req.headers['host'].split(':')[0];
+    
+    // Make sure that when we redirect, it's to the port the proxy is running on
+    if (listen.port !== 443 && listen.port !== 80) {
```
minrk: does this make sense if listen.port is 80?

rgbkrk: Zero sense whatsoever. The only reason I put this in here was a check for non-standard ports. If they're trying to use HTTPS on port 80, I'm currently shuddering. Perhaps some additional warning messages here?

minrk: Maybe just check 443 and ignore https on 80? Does it make sense to have a sanity check to raise if redirectPort is given and sslCert is not specified?

rgbkrk: Yeah, I think a sanity check for certs is wise here. I'll add that in.

rgbkrk:
Cert sanity check in. Wasn't sure if you preferred `process.exit` or `throw new Error(...)` for the main wrapper script. Just let me know either way.

minrk:
I'm not sure. Probably process.exit so there's no traceback, since it's a script, not a library.

bin/configurable-http-proxy
```diff
@@ -25,6 +25,7 @@ args
     .option('--api-ssl-key <keyfile>', 'SSL key to use, if any, for API requests')
     .option('--api-ssl-cert <certfile>', 'SSL certificate to use, if any, for API requests')
     .option('--default-target <host>', 'Default proxy target (proto://host[:port]')
+    .option('--redirect-to-https-from-port <redirect-port>', 'Redirects traffic from redirect port to the port')
```
minrk: I think it's okay to call this `--redirect-port`, and then clarify in the help string that it's `Redirect http requests on this port to the server on https`.

rgbkrk: sounds good

rgbkrk:
Thanks Min, addressed the option name change.

PR #34, opened 2015-10-02
add --error-path

minrk:
for serving html error pages from disk

and use it for *slightly* less basic default error pages


rgbkrk:
This is awesome, thanks @minrk! Same with error target. :)

PR #37, opened 2015-10-02
use bumpversion

minrk:
and set current to 0.5.0-dev

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,21 @@
+[bumpversion]
+current_version = 0.4.1-dev
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z0-9]+))?
+tag_name = {new_version}
+allow_dirty = True
+commit = True
+tag = True
+serialize = 
+	{major}.{minor}.{patch}-{release}
+	{major}.{minor}.{patch}
+
+[bumpversion:file:bin/configurable-http-proxy]
+
+[bumpversion:file:package.json]
+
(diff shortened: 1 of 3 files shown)


rgbkrk:
What's the opposition to `npm version` followed by `npm publish`. The argument to  `npm version` can be major, minor, or patch - it bumps the version and tags for you.

minrk:
@rgbkrk Not really, though I do require that tags are `X.Y.Z`, not `vX.Y.Z`. Can `npm version` be told how to format tags?

rgbkrk:
If using [`npm dist-tag`](https://docs.npmjs.com/cli/dist-tag) directly you can, but they ominously say not to create tags beginning with v:

> Tags that can be interpreted as valid semver ranges will be rejected. For example, v1.4 cannot be used as a tag, because it is interpreted by semver as >=1.4.0 <1.5.0. See https://github.com/npm/npm/issues/6082.

> The simplest way to avoid semver problems with tags is to *use tags that do not begin with a number or the letter v*.

minrk:
dist-tag doesn't write git tags, though, does it? `npm version` seems to unconditionally make tags with `v`. bumpversion, as is the Python way, lets you actually tell it what to do. npm, as is the Javascript way, seems to unconditionally do what I don't want it to, with no ability to tell it to do the right thing.

rgbkrk:
Generally though, you can just override `npm version`:

```
"scripts": {
  "preversion": "npm test",
  "version": "bumpversion <args>",
  "postversion": "git push && git push --tags"
}
```

minrk:
Ah, sure, in which case `npm version` offers nothing at all.

PR #40, opened 2015-12-02
Add a Swagger Spec for the API subset required by JupyterHub

### LINKED ISSUE #39: Add a Swagger (or similar?) spec for the REST API
This would allow creation of alternative proxying setups (See https://github.com/jupyter/jupyterhub/issues/109 for a possible use case) that 'conform to the spec'.

yuvipanda:
Fixes #39 

Not sure exactly if this belongs here - it's only the subset that's strictly required for jupyterhub
to work properly.

### DIFF
--- a/swagger-spec.yaml
+++ b/swagger-spec.yaml
@@ -4,7 +4,14 @@ info:
   description: The REST API JupyterHub requires from a compatible Configurable HTTP Proxy implementation
   version: 0.0.1
 schemes:
-  - https
+  - http
+securityDefinitions:
+  token:
+    type: apiKey
+    name: Authorization
+    in: header
+security:
+  - token: []
 basePath: /api/routes
 produces:
   - application/json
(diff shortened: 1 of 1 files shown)


minrk:
If the goal is to document what JupyterHub requires and nothing else, it wouldn't belong here. But I think we should have a swagger doc here for this. If we want to keep track of the subset that JupyterHub relies upon, we can make a minimal reference on the JupyterHub repo that points to the complete spec here (e.g. JupyterHub uses Configurable HTTP Proxy's API, but does not require host-routing support).

yuvipanda:
Ok, so only the host support is something that CHP has but jupyter doesn't use, right?

yuvipanda:
Probably also need to add info about the Authorization header.

yuvipanda:
Rebased, squashed them around - is generic enough now, I think.

rgbkrk:
Awesome!

PR #42, opened 2015-12-15
add ConfigProxy.proxy_request event

minrk:
and make ConfigProxy an EventEmitter along the way.

Callers may modify the request on its way to upstream by registering a listener on this event

cc @grantlouisherman

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -13,6 +13,7 @@ var http = require('http'),
     https = require('https'),
     fs = require('fs'),
     path = require('path'),
+    EventEmitter = require('events').EventEmitter,
     httpProxy = require('http-proxy'),
     log = require('winston'),
     util = require('util'),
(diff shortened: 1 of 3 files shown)


lib/configproxy.js
```diff
@@ -337,6 +340,7 @@ ConfigurableProxy.prototype.handle_proxy = function (kind, req, res) {
         this.handle_proxy_error(404, kind, req, res);
         return;
     }
+    this.emit("proxy_request", req, res);
```
rgbkrk: Care must be taken with how users work with this event, since it can affect all listeners and the proxy itself.

minrk: That's true, but it's also how the whole proxy works - listening of NHP events to set up the proxying. Is there anything we should be doing differently? try/catch?

rgbkrk: Nah, more like warning signs went off in my head if I was to use this. Not real sure how to document it either.

rgbkrk:
:+1: :shipit:

PR #44, opened 2015-12-30
more SSL options

minrk:
- `--ssl-ciphers` for specifying cipher string
- `--[api]-ssl-ca` for specifying CA

Default for `--ssl-ciphers` excludes RC4, which can be overridden with `--ssl-allow-rc4`.

Default behavior with letsencrypt certs [gets an A](https://www.ssllabs.com/ssltest/analyze.html?d=chp.minrk.net) on SSL labs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,17 +45,21 @@ in the proxy table:
 
     -h, --help                       output usage information
     -V, --version                    output the version number
-    --ip <n>                         Public-facing IP of the proxy
+    --ip <ip-address>                Public-facing IP of the proxy
     --port <n>                       Public-facing port of the proxy
     --ssl-key <keyfile>              SSL key to use, if any
     --ssl-cert <certfile>            SSL certificate to use, if any
+    --ssl-ca <ca-file>               SSL certificate authority, if any
+    --ssl-ciphers <ciphers>          `:`-separated ssl cipher list. Default excludes RC4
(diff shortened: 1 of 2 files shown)


rgbkrk:
Awesome, that's great.

`master` is currently passing, I'm not sure of the error reported here.

minrk:
@rgbkrk any other SSL options you can think of that aren't exposed that should be?

rgbkrk:
I'd rely on whatever the [Mozilla SSL Configuration Generator](https://mozilla.github.io/server-side-tls/ssl-config-generator/) hands you.

The only option I'm curious about is choosing protocols, we'd want explicitly only `TLSv1.1` and `TLSv1.2` for fully modern, and you can back pedal adding `TLSv1` followed by `SSLv3` if older support is necessary.

If you want the score any higher on SSL Labs, we'd have to generate at runtime (or allow for setting) the Diffie Hellman parameters. Config flag for that makes sense, though I don't think it has to be in this PR.

minrk:
ok, thanks. I think node already sets sensible defaults for TLS. Since I'm not clear on what the best way to express that option as a flag would be, I'll leave it as-is, for now.

rgbkrk:
Well, similar to how you'd expect the certs as files, same goes for dhparam.pem. If they don't specify it, we'd rely on the defaults.

minrk:
Added dhparam in #47

PR #47, opened 2016-01-02
allow specifying Diffie-Helman parameters via `--ssl-dhparam`

minrk:
For perfect-forward secrecy.

@rgbkrk does it make sense to generate this on-demand if unspecified, rather than require loading it from a file? I'm not certain what's in a dhparam file.


minrk:
Also, is dhparam unused when using ECDHE instead of DHE? Is there a need for it?

rgbkrk:
It can take hours to generate at higher key sizes, I wouldn't generate them automatically.

minrk:
ok, then. I'll cut a release with this shortly.

rgbkrk:
> Also, is dhparam unused when using ECDHE instead of DHE? Is there a need for it?

This I'm not sure of.

PR #51, opened 2016-03-29
Add basic statsd metrics

yuvipanda:
- Sends very basic statsd metrics to a configurable endpoint
- When statsd is not configured, nothing happens.

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -44,6 +44,9 @@ args
     .option('--no-include-prefix', "Don't include the routing prefix in proxied requests")
     .option('--insecure', "Disable SSL cert verification")
     .option('--host-routing', "Use host routing (host as first level of path)")
+    .option('--statsd-host <host>', 'Host to send statsd statistics to')
+    .option('--statsd-port <port>', 'Port to send statsd statistics to', parseInt)
+    .option('--statsd-prefix <prefix>', 'Prefix to use for statsd statistics')
     .option('--log-level <loglevel>', 'Log level (debug, info, warn, error)', 'info');
 
 args.parse(process.argv);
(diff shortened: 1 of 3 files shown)


yuvipanda:
You can setup a real statsd instance to test by following the following:

1. Clone https://github.com/etsy/statsd
2. Run 'npm install' inside the clone
3. Create a config file with the following contents: `{ backends: [ "./backends/console" ] }`
4. Save config file as `config.js`
5. Start statsd with `node stats.js config.js`

Then pass --statsd-host as localhost to chp when starting and perform some operations - the statsd server will output the counters and what not to the console.

yuvipanda:
The test failures seem unrelated.

bin/configurable-http-proxy
```diff
@@ -44,6 +44,9 @@ args
     .option('--no-include-prefix', "Don't include the routing prefix in proxied requests")
     .option('--insecure', "Disable SSL cert verification")
     .option('--host-routing', "Use host routing (host as first level of path)")
+    .option('--statsd-host <host>', 'Host to send statsd statistics to')
+    .option('--statsd-port <port>', 'Port to send statsd statistics to', parseInt)
+    .option('--statsd-prefix <prefix>', 'Prefix to use for statsd statistics')
```
minrk: can we make this a statsd-endpoint with `host[:port][/prefix]`?

yuvipanda: We could, but that's somewhat nonstandard (from my experience, at least!). The prefix is also not a URL prefix, but a prefix to use for the metrics - so if prefix is 'tools.paws.chp' (as it is in my case), the metrics would be tools.paws.chp.request.web, etc. So putting them in URL notation doesn't make sense, IMO.

minrk: Fair enough.

minrk:
Not entirely sure how, but test failures do seem to be caused by this PR. I can reproduce it locally and on Travis, and without this PR both work.

lib/configproxy.js
```diff
@@ -174,6 +175,7 @@ ConfigurableProxy.prototype.add_route = function (path, data) {
     this.routes[path] = data;
     this.trie.add(path, data);
     this.update_last_activity(path);
+    this.statsd.increment('route.add', 1);
```
minrk: This (and other calls) should be protected by `if (this.statsd)`.

yuvipanda: Nope - if you look at bin/configurable-http-proxy, I basically setup a stub statsd object that does nothing. This allows me to not need to guard against a null statsd everywhere.

minrk: This is causing the failing tests. CHP can be instantiated as a library via js API, in which case the bin script isn't involved.

minrk: So either that default no-op needs to be moved to the ConfigProxy constructor, or an undefined/null statsd should be handled (I would guess handling null statsd is cheaper than calling no-op functions, but I'm not sure).

yuvipanda: I've moved the noop to inside the constructor. I think the nodejs JIT will optimize away the empty calls.

yuvipanda:
If I do 'npm test' now it passes for me.

yuvipanda:
Tests pass now! \o/

yuvipanda:
Outside of discussion about the naming of the metrics, I think this is ok to merge now.

rgbkrk:
I love this and the names look great. Would love to see us use this on try.jupyter.org's config proxies.

yuvipanda:
This will need https://github.com/jupyter/jupyterhub/pull/496 to be used from installs where JupyterHub is spawning CHP.

PR #54, opened 2016-04-01
Emit statsd metrics for 5xx error codes as well

yuvipanda:


lib/configproxy.js
```diff
@@ -310,6 +310,7 @@ ConfigurableProxy.prototype.handle_proxy_error = function (code, kind, req, res)
 
     var proxy = this;
     log.error("%s %s %s", code, req.method, req.url);
+    that.statsd.increment('requests.' + code, 1);
```
minrk: s/that/this/

yuvipanda: fixed

PR #55, opened 2016-04-27
handle args.ip = '*'

minrk:
interpret as alias for all-interfaces, recommend correct spelling of the same option.

closes jupyterhub/jupyterhub#545


rgbkrk:
Thus begins my quest to retrigger Travis builds until this passes.

PR #57, opened 2016-04-27
Test setting the default target on startup.

rgbkrk:
Mostly looking to increase branch coverage here (`nyc` showed this as not covered so I tested it).

I've been letting whitespace changes sneak through from my editor. If you want me to rebase with only the actual test, I can do that.

### DIFF
--- a/test/proxy_spec.js
+++ b/test/proxy_spec.js
@@ -5,6 +5,8 @@ var util = require('../lib/testutil');
 var request = require('request');
 var WebSocket = require('ws');
 
+var ConfigurableProxy = require('../lib/configproxy').ConfigurableProxy;
+
 describe("Proxy Tests", function () {
     var port = 8902;
     var test_port = port + 10;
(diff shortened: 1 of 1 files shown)


minrk:
I don't care too much about the whitespace, but this conflicts with #56, which I've merged. That wouldn't have happened if you hadn't committed those whitespace changes :)

It does bother me a lot that this is atom's default behavior. I [tried to fix it](https://github.com/atom/whitespace/pull/22) long ago, but they made it clear that they want to impose their opinions on everyone by default. To me, it's enough of a dick move by the devs that I would never recommend atom to any new developers.

rgbkrk:
Ha, I'll rebase soon enough.

rgbkrk:
Done deal.

minrk:
I was hoping for the coverage PR to give you a nice bonus :cake: for this PR, but alas codecov still seems to be reporting a false 100% coverage (not the most useful).

PR #59, opened 2016-04-28
track test coverage

minrk:


minrk:
Weirdly, codecov is [showing 100% coverage](https://codecov.io/gh/jupyterhub/configurable-http-proxy/tree/f8ea6d5a0a4c41486ea09935486922e9cfb58466/lib), but the coverage report on travis [is not](https://travis-ci.org/jupyterhub/configurable-http-proxy/jobs/126348531).

rgbkrk:
Odd. I used nyc locally last night and definitely didn't have 100% coverage.

minrk:
Istanbul isn't reporting 100% on Travis, but codecov is. Not sure what's up, there.

stevepeak:
Hey guys, fix is on its way. There was an issue with the `coverage.json`. Which has been somewhat of a nightmare to deal with. The coverage stated hit from line X to Z then a miss on Y... and the Y is accurate.

Deploying this shortly 👍

minrk:
@stevepeak thanks! I've also opened #61 which switches to nyc, which also fixes the issue. I suppose the only relevant difference is that it doesn't create the coverage.json.

PR #60, opened 2016-04-28
test with mocha instead of jasmine

minrk:
maybe the multiple runs is what confuses codecov

Let's wait for the codecov report before merging, in case this doesn't change anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,10 @@
     "winston": "~0.8"
   },
   "devDependencies": {
-    "jasmine-node": "~2",
+    "chai": "^3.5.0",
+    "chai-subset": "^1.2.2",
     "jshint": "*",
+    "mocha": "^2.4.5",
     "nyc": "^6.4.0",
     "request": "~2",
     "ws": "~0.4"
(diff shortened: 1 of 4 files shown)


minrk:
Nope, that wasn't it. Trying another thing...

minrk:
still no. Trying a different test runner, perhaps.

minrk:
@rgbkrk this doesn't fix the codecov issue, but do you have a preference for mocha vs jasmine (or another test runner)? You have a lot more js experience than I do at this point.

minrk:
Switching to nyc with lcov reports seems to fix the issue.

minrk:
#61 just does the nyc switch fixing codecoov, so this PR is now switching the runner to mocha. I'm not sure if I want to do that or not, but the work is done, and I'll leave it here.

rgbkrk:
I tend to use `mocha` though I'm not really opinionated there. I've adapted for jasmine and tape as well.

minrk:
Leaving this with jasmine for now. I don't see an advantage to mocha, but we can pick this up as a starting point if that changes.

PR #63, opened 2016-05-03
add `--pid-file`

minrk:
writes PID to a file

Also handle uncaught exceptions and sigint

- exit on sigint, ensuring PID cleanup happens
- log, don't die on uncaught exceptions once the server has started


rgbkrk:
👍

PR #67, opened 2016-07-15
expose `--ssl-protocol`

minrk:
for specifying TLSv1_2, etc.

cc @tmtabor


bin/configurable-http-proxy
```diff
@@ -114,6 +115,10 @@ if (args.sslKey || args.sslCert) {
     if (args.sslDhparam) {
         options.ssl.dhparam = fs.readFileSync(args.sslDhparam);
     }
+    if (args.sslProtocol) {
+      options.ssl.secureProtocol = args.sslProtocol + '_method';
+    }
+    console.log(options.ssl);
```
rgbkrk: Did you mean to leave this `console.log` in?

minrk: D'oh, nope. Fixed.

PR #69, opened 2016-07-30
fix forward args

### LINKED ISSUE #68: noXHeaders -> noXForward ?
Should the argument or variable name change here ?
https://github.com/jupyterhub/configurable-http-proxy/blob/master/bin/configurable-http-proxy#L181

quasiben:
resolves #68


rgbkrk:
:smile: Thanks Min.

PR #73, opened 2016-09-07
Protocol/host rewrite

stefanoborini:
Exports the auto-rewrite and protocol-rewrite options from http-proxy to the command line.
Also verifies that the feature actually works.


minrk:
Thanks!

PR #81, opened 2016-09-28
Separate storage mechanism for routes

pseudomuto:
@minrk 

# The Problem

Currently the code that handles persistence of routes is embedded directly in the proxy code. This makes it difficult to choose a different storage method (e.g. redis, memcached, db, etc.).

This issue becomes apparent when you're running multiple JupyterHub instances behind nginx (or some other proxy). In this scenario, each JH instance will spawn their own proxy and maintain their own route table (in the node proc).

When a request comes in at `/user/xyz`, depending on the host that nginx routes to, the local proxy will either have the route, or not. 

😢 

Initially I patched the JH Proxy object to update all routes and lowered the last_activity_timeout param so it would happen often, but this seemed like a total hack and doesn't always work (still a small window of time where the error can occur).

# The Solution

Ideally, storage should be configurable/pluggable. I'm thinking something like passing a config option to change the persistence method or something.

In order to keep this PR small and reviewable, I've simply moved the storage from `configproxy.js` into it's own `store.js` module. The trie is still used, and all storage is still in memory (i.e. nothing changes).

# Next Steps

If this seems reasonable, then I'll add implementations for other providers in a separate PR.

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -290,7 +290,7 @@ ConfigurableProxy.prototype.update_last_activity = function (prefix) {
     // note last activity in routing table
     if (this._routes.hasRoute(prefix)) {
         // route may have been deleted with open connections
-        this._routes.get(prefix).last_activity = new Date();
+        this._routes.update(prefix, { "last_activity": new Date() });
     }
     timer.stop();
 };
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -208,9 +208,9 @@ ConfigurableProxy.prototype.remove_route = function (path, cb) {
     var routes = this._routes;
 
     routes.hasRoute(path, function (result) {
-      if (result) {
-        routes.remove(path, cb);
-      }
+        if (result) {
+            routes.remove(path, cb);
+        }
     });
 };
 
@@ -221,7 +221,7 @@ ConfigurableProxy.prototype.get_routes = function (req, res) {
     var inactive_since = null;
     if (parsed.query) {
         var query = querystring.parse(parsed.query);
-        
+
         if (query.inactive_since !== undefined) {
             var timestamp = Date.parse(query.inactive_since);
             if (isFinite(timestamp)) {
@@ -238,15 +238,14 @@ ConfigurableProxy.prototype.get_routes = function (req, res) {
         var results = {};
 
         if (inactive_since) {
-            var keys = Object.keys(routes).filter(function (key) {
-                return routes[key].last_activity < inactive_since;
+            Object.keys(routes).forEach(function (path) {
+                if (routes[path].last_activity < inactive_since) {
+                    results[path] = routes[path];
+                }
             });
-
-            keys.forEach(function (key) { results[key] = routes[key]; });
         } else {
             results = routes;
         }
-
 
         res.write(JSON.stringify(results));
         res.end();
@@ -300,8 +299,15 @@ ConfigurableProxy.prototype.target_for_req = function (req, cb) {
 
     this._routes.getTarget(base_path + decodeURIComponent(req.url), function (route) {
         timer.stop();
-        var result = route ? { prefix: route.prefix, target: route.data.target } : null;
-        cb(result);
+        if (route) {
+            cb({
+                prefix: route.prefix,
+                target: route.data.target
+            });
+            return;
+        }
+
+        cb(null);
     });
 };
 
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -169,7 +169,7 @@ function ConfigurableProxy (options) {
     }
 
     // proxy requests separately
-    var proxy_callback = log_errors(that.handle_proxy_web);
+    var proxy_callback = log_errors(this.handle_proxy_web);
     if ( this.options.ssl ) {
         this.proxy_server = https.createServer(this.options.ssl, proxy_callback);
     } else {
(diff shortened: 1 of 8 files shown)


pseudomuto:
CI failure seems to be flaky tests. [This one](https://travis-ci.org/jupyterhub/configurable-http-proxy/jobs/163480191) for instance passed on the retry, but failed initially

rgbkrk:
Great, thanks for bringing this up! I agree that we need to have a way to persist routes reliably across many nodes.

minrk:
Thanks! I think it's fine to add this functionality. I'm not sure the best way to go about it. The things I'd be concerned about are:

1. what's the best way to add providers? I don't want CHP to grow too much, and adding a provider in general shouldn't have to be a PR against this repo, and I don't love adding dependencies that people don't use.
2. I would prefer to avoid route lookup becoming an async operation, but it would presumably have to be for any external storage, and therefore also become an immediately-evaluated async API for the in-memory case, as well.

I don't know how to achieve 1 best, given how npm bundles dependencies. I'd also be worried about growing optional dependencies, but at least the redis and memjs clients aren't problematic (other memcache clients appear to be).

pseudomuto:
@minrk I agree with all of that!

I'm just working on making the storage async as we speak. As for adding deps/providers in this repo, I'm not entirely sure what the best move would be just yet. One of those "I'll burn that bridge when I get there" kinda things :smile:

Maybe [peer dependencies](https://nodejs.org/en/blog/npm/peer-dependencies/) could be used somehow?

minrk:
I don't think the tests have been flaky recently, so that seems like there might be something amiss in the changes here, especially since every run is failing in the same way, and a push from master [just now](https://travis-ci.org/minrk/configurable-http-proxy/builds/163500082) is all green. This PR does pass all tests when I run on my laptop, though, which makes debugging harder.

Re: peerDependencies, maybe! I'm far from an expert on npm packaging, and mostly get frustrated the more I try to solve nontrivial problems with it. My first impression is that peerDeps would solve it well where configproxy is used *as a library*, but may not work well for the `npm install -g`-style application used most often. I would **love** for someone else to figure out the best way to deal with that. Let us know what you figure out!

pseudomuto:
It's 🍏 !!!

I've updated the code to now use an async API for storage. Not gonna lie, it was a little rough. This is much bigger than I originally intended, so Iet me apologize in advance :smile:

> I don't think the tests have been flaky recently, so that seems like there might be something amiss in the changes here

This was indeed my changes. The first was actually a lint error, and the 0.x versions didn't support `Object.assign`.

minrk:
You're right, this is really huge! The async changes are unfortunate, but I guess there's no way around it.

It apparently also contains a bunch of aesthetic changes? I wish that didn't happen, as the PR is big enough without those, especially since in most cases I prefer the before to the after. Was there a reason for those? Or did you use an auto-formatter?

lib/configproxy.js
```diff
@@ -222,20 +232,25 @@ ConfigurableProxy.prototype.get_routes = function (req, res) {
...
-    res.write(JSON.stringify(routes));
-    res.end();
-    this.statsd.increment('api.route.get', 1);
+
+    this._routes.getAll(function (routes) {
+        var results = {};
+
+        if (inactive_since) {
+            var keys = Object.keys(routes).filter(function (key) {
+                return routes[key].last_activity < inactive_since;
```
minrk: The previous implementation was more efficient because it iterated over the routes only once, rather than twice.

lib/configproxy.js
```diff
@@ -249,47 +264,63 @@ ConfigurableProxy.prototype.post_routes = function (req, res, path, data) {
...
-    if (route) {
-        return {
-            prefix: route.prefix,
-            target: route.data.target,
-        };
-    }
+
+    this._routes.getTarget(base_path + decodeURIComponent(req.url), function (route) {
+        timer.stop();
+        var result = route ? { prefix: route.prefix, target: route.data.target } : null;
```
minrk: It would be nice to leave the less-dense format above.

test/api_spec.js
```diff
@@ -115,91 +137,83 @@ describe("API Tests", function () {
...
-                    '/yesterday': true,
-                    '/today': true
-                },
-            },
+            { name: 'long ago', since: long_ago, expected: {} },
+            { name: 'an hour ago', since: hour_ago, expected: { '/yesterday': true } },
+            { name: 'the future', since: hour_from_now, expected: { '/yesterday': true, '/today': true } }
         ];
 
-        r.get(api_url + "?inactive_since=endoftheuniverse", function (error, res, body) {
```
minrk: This check was removed. Can it come back?

test/api_spec.js
```diff
@@ -115,91 +137,83 @@ describe("API Tests", function () {
...
-                name: 'the future',
-                since: hour_from_now,
-                expected: {
-                    '/yesterday': true,
-                    '/today': true
-                },
-            },
+            { name: 'long ago', since: long_ago, expected: {} },
+            { name: 'an hour ago', since: hour_ago, expected: { '/yesterday': true } },
+            { name: 'the future', since: hour_from_now, expected: { '/yesterday': true, '/today': true } }
```
minrk: FWIW, I would prefer this kind of style change to not happen (before is better than after, for me).

test/api_spec.js
```diff
@@ -115,91 +137,83 @@ describe("API Tests", function () {
...
-        util.add_target(proxy, '/today', port+1);
 
-        var now = new Date();
-        var yesterday = new Date(now.getTime() - (24 * 3.6e6));
-        var long_ago = new Date(1);
-        var hour_ago = new Date(now.getTime() - 3.6e6);
+        var now           = new Date();
+        var yesterday     = new Date(now.getTime() - (24 * 3.6e6));
+        var long_ago      = new Date(1);
+        var hour_ago      = new Date(now.getTime() - 3.6e6);
```
minrk: I haven't done aligned assignments generally, and would prefer them not to be applied to existing code without any other changes. Not a big deal, just something for the future.

test/api_spec.js
pseudomuto: I just moved it to it's own test (above)

pseudomuto:
Updated to remove all of the style-only changes (align assignments, etc.). I've also rebase on the latest master with @rgbkrk's changes

test/api_spec.js
minrk: Sorry I missed that, thanks!

lib/store.js
```diff
@@ -62,7 +62,12 @@ function MemoryStore () {
     },
     update: {
       value: function (path, data, cb) {
-        Object.assign(routes[path], data);
```
rgbkrk: We could probably go back to this to simplify the logic in a later PR.

willingc:
Thank you @pseudomuto 😎

PR #82, opened 2016-09-29
Only support LTS and greater releases of node

rgbkrk:
To use newer features of JavaScript and Node.js that are useful _and_ performant (like native `Object.assign`, built-in `Map`), we should stick to more recent versions of Node.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ node_js:
   - "5"
   - "4"
   - "0.12"
-  - "0.11"
-  - "0.10"
 install:
   - npm install -g codecov
   - npm install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: node_js
 sudo: false
 node_js:
+  - "6"
   - "5"
   - "4"
-  - "0.12"
 install:
   - npm install -g codecov
   - npm install


pseudomuto:
Nice!

Might need to remove `0.12` as well. I found the issue by switching to `0.12` with nvm and running `npm test test/store_spec.js`

rgbkrk:
I did it based on the `[LTS schedule](https://github.com/nodejs/LTS)`, so even though I don't use 0.12 I left it in.

We can certainly cut `0.12` too though, I'm in favor.

willingc:
Whatever we ultimately decide re: `0.12`, we should probably link to https://github.com/nodesource/distributions to give guidance to those users on slow moving LTS versions of Ubuntu or Fedora.

rgbkrk:
I find it amusing that the tests are intermittently failing only on `0.12`.

I agree with linking to node source's guidance with respect to LTS versions of Ubuntu and Fedora.

willingc:
Unless @minrk objects, I say remove 0.12 and we can link in the readme to the node guidance re: older versions.

minrk:
👍 I've been meaning to add 6 for a while, but haven't had time to look at why the tests fail. It seems to be testing-related, as I'm pretty sure I've run CHP on node 6 without issues.

PR #83, opened 2016-09-30
query string is an object, but also not an object, on node6

minrk:
hasOwnProperty is not defined, even though `typeof q` is object.

This should get tests passing on node 6


rgbkrk:
Odd.

minrk:
Turns out this was [deliberate](https://github.com/nodejs/node/issues/7223) though there doesn't appear to be any public reasoning behind it.

PR #86, opened 2016-10-01
Add proxy-timeout option

ashishdahiya:
http-proxy lib provides a [proxyTimeout option](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L122) that allows us to set a timeout on the outgoing socket connection to the target. This timeout is very effective when the upstream target does not respond within an expected time.

I had wasted a few hours searching for this option. Documenting this option can save others a significant amount of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ matching route is found in the proxy table:
     --statsd-port <port>             Port to send statsd statistics to
     --statsd-prefix <prefix>         Prefix to use for statsd statistics
     --log-level <loglevel>           Log level (debug, info, warn, error)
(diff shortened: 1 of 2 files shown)


minrk:
While this adds a CLI option, it doesn't hook it up to anything, so it won't have any effect. Do you want to add `options.proxyTimeout = args.proxyTimeout`?

ashishdahiya:
You are correct. I missed that. I have added it now. PTAL.

minrk:
Thanks!

PR #87, opened 2016-10-14
Added label system [WIP]

stefanoborini:
This patch adds a label to a setup route, so that we can then use the label to remove routes associated to that label.

Work in progress as I want to hear your feedback (and I'll also have to update the documentation with the new API)

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -230,7 +230,7 @@ ConfigurableProxy.prototype.remove_route = function (path, cb) {
                     var paths = that.labels[data.label];
                     var idx = paths.indexOf(path);
 
-                    if (idx != -1) {
+                    if (idx !== -1) {
                         paths.splice(idx, 1);
                     }
                 }
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -104,6 +104,7 @@ function ConfigurableProxy (options) {
     this._routes = store.MemoryStore();
     this.auth_token = this.options.auth_token;
     this.includePrefix = options.includePrefix === undefined ? true : options.includePrefix;
+    this.labels = {};
     this.host_routing = this.options.host_routing;
     this.error_target = options.error_target;
     if (this.error_target && this.error_target.slice(-1) !== '/') {
@@ -145,6 +146,13 @@ function ConfigurableProxy (options) {
             get : bound(this, authorized(this.get_routes)),
             post : json_handler(bound(this, authorized(this.post_routes))),
             'delete' : bound(this, authorized(this.delete_routes))
+        } ],
+        [ /^\/api\/labels\/?$/, {
+            get : bound(this, authorized(this.get_labels)),
+        } ],
+        [ /^\/api\/labels\/?(.*)$/, {
+            get : bound(this, authorized(this.get_label)),
+            'delete' : bound(this, authorized(this.delete_label))
         } ]
     ];
 
@@ -194,6 +202,13 @@ ConfigurableProxy.prototype.add_route = function (path, data, cb) {
 
     var that = this;
 
+    if (data.hasOwnProperty("label")) {
+        if (!(data.label in this.labels)) {
+            this.labels[data.label] = [];
+        }
+        this.labels[data.label].push(path);
+    }
+
     this._routes.add(path, data, function () {
       that.update_last_activity(path, function () {
         if (typeof(cb) === "function") {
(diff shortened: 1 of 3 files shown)


yuvipanda:
I'm curious what the use case for this is?

stefanoborini:
@yuvipanda I am building a system that uses jupyterhub to handle multiple docker containers for multiple users. Some users are administrators and are authorized to stop other users' containers. However, in addition to stop them, they also need to remove the url from the reverse proxy. Problem is, the administrative program cannot know which url has been setup for the user, because it depends on the spawner decision of the base url, and the spawned program convention for the created url.

so the administrative program has too much variability to decide what url to remove from the revproxy. _but_ I have my docker container id, and when I start them, I can add the url to the revproxy with that id as a label. When the admin program has to stop the container, it takes its id, stops the container, then asks the revproxy to remove all urls that are associated to that id.

yuvipanda:
It's awesome you're working on multiple containers per user! Are they just multiple notebook containers? That sounds similar to https://github.com/jupyterhub/jupyterhub/issues/766, so I'm wondering if this work is being done inside jupyterhub itself or outside it? If outside, wouldn't you need to make JupyterHub itself be aware of what's going on by calling it's API, in which case you can use that to track which routes need to be removed?

In general, I'm slightly wary of adding more complexity to the proxy - and if anything would like to make it simpler. https://github.com/jupyterhub/jupyterhub/issues/502 for example - keeping this state in the proxy means it can not scale horizontally at all. While the labels system probably doesn't add any such constraints, it still does make life difficult for other implementations of CHP to happen (I've been working on an nginx based one, for example). 

As such, I'm hoping I can convince you that you don't need this, and this can be implemented in a different fashion elsewhere :) Is there a repo or design document online somewhere I can look at to see what you're trying to do?

yuvipanda:
I'd also like someone (perhaps @minrk?) to clarify how closely tied to JupyterHub CHP is. Currently it seems to be in a bit of an uncanney valley - it already has some jupyterhub specific stuff (it's tracking which users are registered explicitly, for example), but the rest of the interfaces are a bit too generic. This makes it not too useful for being a generic configurable proxy (and there are plenty of others that do that well - vulcand, traefic, etc) and causes it to have some mismatch with JupyterHub's model.

We should either decide that it's going to be completely generic, and if so figure out what to do with 'user' fields, or decide it's JupyterHub specific, which would make situations like this easier to solve.

stefanoborini:
> Are they just multiple notebook containers?

No, it's a bit more complicated. See https://github.com/simphony/simphony-remote

> If outside, wouldn't you need to make JupyterHub itself be aware of what's going on by calling it's API

I use JupyterHub only for authentication and spawning. Once that's out of the way, I take over the interaction. My spawned "singleuser" talks with the reverse proxy to add more urls.

> in which case you can use that to track which routes need to be removed?

In principle, I could, but that would require me to track it on a shared database, so that user A adds the url to this database when it starts a container, and user B (admin) can use the database to remove the url when it turns off the container. This patch makes the reverse proxy that database, in a sense. So yes, there is a different strategy, but it feels more natural to put it here as a native functionality.

> As such, I'm hoping I can convince you that you don't need this

well, I do, I don't necessarily need it _here_, but it felt like a natural choice. 

> Is there a repo or design document online somewhere I can look at to see what you're trying to do?

The url I gave above. Any feedback for anything you might think it's badly implemented is welcome. I am not very experienced in web programming, but I am building up some muscles.

minrk:
@stefanoborini you can already store data on the routes, so you can do:

```
POST /api/routes/prefix
{ "target": "http://endpoint",
  "label": "name"
}
```

What do you get from adding a whole labels API that you don't already have by storing label data on the routes themselves?

@yuvipanda  while CHP exists for the purpose of deploying JupyterHub, it's a generic tool, useful for other things (e.g. tmpnb, binder, etc.). Due to that, I would argue against any changes that adversely affect JupyterHub, but not necessarily other generically useful features that JupyterHub doesn't happen to use.

stefanoborini:
@minrk The problem is that even if I add the label as data, I can't search for it. I need to remove by label, without knowing the url it's associated to.

yuvipanda:
On Sun, Oct 16, 2016 at 3:28 AM, Stefano Borini <notifications@github.com>
wrote:

> in which case you can use that to track which routes need to be removed?
>
> In principle, I could, but that would require me to track it on a shared
> database, so that user A adds the url to this database when it starts a
> container, and user B (admin) can use the database to remove the url when
> it turns off the container. This patch makes the reverse proxy that
> database, in a sense. So yes, there is a different strategy, but it feels
> more natural to put it here as a native functionality.
>
Just to respond to this - CHP loses all state when it restarts, and JHub
has to repopulate it (from its own state).  So persistent state that needs
to survive restarts needs to be stored somewhere anyway...

stefanoborini:
> Just to respond to this - CHP loses all state when it restarts, and JHub
has to repopulate it (from its own state).  So persistent state that needs
to survive restarts needs to be stored somewhere anyway...

Well that would not be a problem. If CHP restarts it loses all the mappings, and that's fine. My code is made so that if that occurs, the containers stay up, and the url is re-added as soon as it views it.

minrk:
Since there's already a mechanism for storing data per path, adding a second one for one particular type of data doesn't seem like the right thing to do. What if you added the ability for DELETE to to operate based on the data stored on routes, instead, e.g.:

```
DELETE /api/routes?label=value
```

deletes routes where `data.label === value`?

stefanoborini:
After gitter discussion, I will add query functionality to DELETE in the form of /api/routes/?label=foobar. Closing this one.

PR #88, opened 2016-10-17
Ability to query while extracting a specific data entry

stefanoborini:
Filters the results according to the passed query parameters. With /api/routes/?foo=bar it will extract only those entries whose data contains foo=bar

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -19,6 +19,13 @@ var http = require('http'),
     querystring = require('querystring'),
     store = require('./store.js');
 
+var InvalidQuery = function (message) {
+    return {
+        name: "InvalidQuery",
+        message: message
+    };
+};
+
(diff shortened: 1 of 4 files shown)


willingc:
@minrk Is this something that can be merged or is there additional work needed?

stefanoborini:
@willingc It was my understanding that it was not really needed (one could use alternative strategies), but I left the PR open in case it was topic of discussion

willingc:
Thanks for the perspective @stefanoborini. Should I close the PR and label it as Reference for the benefit of others in the future?

stefanoborini:
As far as I am concerned, it can be closed.

willingc:
Closing this and marking it as reference.

Thank you @stefanoborini for the PR and your ongoing insights and support on JupyterHub. :cake:

PR #91, opened 2016-12-09
fix closure reference when serving custom error pages

minrk:
this != that!

cf https://github.com/jupyterhub/jupyterhub/issues/900


minrk:
thanks!

rgbkrk:
Yeah, nodejs has been having trouble with their hosted binaries in the last 24 hours or so. They're working on fixing it.

PR #94, opened 2017-04-03
Clarify ip all-interfaces warning

### LINKED ISSUE #78: Update warning message ip address
In https://github.com/jupyterhub/configurable-http-proxy/blob/master/bin/configurable-http-proxy#L216, we caution against using '*' as an ip in the docs and in the warning a few lines above yet it is displayed by L216. Perhaps change to '0.0.0.0' or '' or emit a warning such a few lines above.

willingc:
Closes #78 

We could delete the reference to 0.0.0.0 if the warning message seems too long.


rgbkrk:
Thanks!

PR #98, opened 2017-04-06
specify engine requirement in package.json

minrk:
this should get a better error message on installation with unsupported node versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,10 @@
     "request": "~2",
     "ws": "^1.1"
   },
+  "engines": {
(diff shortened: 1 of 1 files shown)


minrk:
Error looks like this:

```
npm ERR! notsup Unsupported
npm ERR! notsup Not compatible with your version of node/npm: configurable-http-proxy@2.1.0-dev
npm ERR! notsup Required: {"node":">= 4.0"}
npm ERR! notsup Actual:   {"npm":"2.15.11","node":"0.12.18"}
```

PR #104, opened 2017-04-19
Docker updates

### LINKED ISSUE #101: Help running the docker container
Requesting just a little bit of documentation for use of the Docker container.

I've tried running the docker container like so:

```
docker run -it -p 80:8000 -p 8001:8001 jupyterhub/configurable-http-proxy --default-target http://localhost:8080
```

But the server running on port 8080 doesn't receive requests, and the proxy reports a refused connection as so:

```
06:26:23.522 - error: [ConfigProxy] Proxy error:  Error: connect ECONNREFUSED 127.0.0.1:8080
    at Object.exports._errnoException (util.js:893:11)
    at exports._exceptionWithHostPort (util.js:916:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1075:14)
```

I've confirmed I can reach the server on port 8080 directly.

I can fix this by changing the default route to the machine's actual IP.

```
docker run -it -p 80:8000 -p 8001:8001 jupyterhub/configurable-http-proxy --default-target http://10.0.0.7:8080
```

If I try to reach the API, I can do a GET, but it doesn't report the default route.

```
› curl http://localhost:8001/api/routes
curl: (52) Empty reply from server
```

POSTing a new route does not work...

```
curl -i -X POST http://localhost:8001/api/routes -d '{"/foo":{"target":"http://localhost:8082"}}'

curl: (52) Empty reply from server

curl http://localhost:8001/api/routes
curl: (52) Empty reply from server
```

I've also tried deriving a container and exposing the 8001 port explicitly, but I don't see any difference in behavior.

### LINKED ISSUE #102: Using node:alpine as base image?
With setting the base image to node:alpine and building the docker container I got image of size 61 MB,
whilst the node:5-slim image has size 209 MB. I am testing Alpine Linux based image in Jupyter tmpnb notebooks on our university server, no problems as for now.
 
My consideration is not only about size. node:alpine has 13 (apk installed) packages, each of them really necessary but "slim" variant (debian based) has 135 packages, some of them very superfluous in this concrete case (in reality, I am debian user for many years...).   

Are there any reasons not to use node:alpine as the base image?

minrk:
- base default docker image on node:6-alpine for reduced size (removes separate alpine Dockerfile)
- Change default api ip in the image to all interfaces,
  so that the API port is accessible via port forwarding, etc.
  without having to also specify `--api-ip` by hand.

closes #101
closes #102


chp-docker-entrypoint
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Wrapper around CHP entrypoint that changes defaults slightly
+# to be more appropriate when run in a container.
+
+if [[ "$@" != *"--api-ip"* ]]; then
+  # Default api-ip to all interfaces in docker,
+  # so that it is accessible to other containers
+  # and when port-forwarding is requested.
+  ARGS="--api-ip=0.0.0.0"
```
willingc: Does the ip address need to be in quotes? `ARGS="--api-ip='0.0.0.0'"`

minrk: it does not

PR #110, opened 2017-06-20
improve logging consistency

minrk:
- log status, path of every API request
- log modifications to routing table at info-level
- log failed authentication details at debug-level

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -35,13 +35,13 @@ function arguments_array (args) {
 function fail (req, res, code, msg) {
     // log a failure, and finish the HTTP request with an error code
     msg = msg || '';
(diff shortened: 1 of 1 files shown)


lib/configproxy.js
```diff
@@ -459,22 +462,47 @@ ConfigurableProxy.prototype.handle_proxy_web = function (req, res) {
...
 };
 
+function log_response(req, res) {
+    // log function called when any response is finished
+    var code = res.statusCode;
+    var log_f = log.info;
+    if (code >= 400) {
+        log_f = log.warn;
+    } else if (code >= 500) {
+        log_f = log.error;
```
willingc: @minrk I'm assuming that you wish to log a warning and an error when code is greater than 500.

PR #118, opened 2017-07-21
Quieter messages for ECONNRESET and ECONNREFUSED

minrk:
These error codes occur for upstream issues, and aren't generally problems in the proxy.

- ECONNREFUSED means the target is not accepting connections (generally because it's not running)
- ECONNRESET means that a connection was made, but didn't receive a response before losing the connection
  (e.g. upstream shutdown before finishing, or didn't respond in a timely fashion)

We were logging stack traces for these messages, which users have reported as possible proxy problems, when they are generally not anything to be concerned about.

closes https://github.com/jupyterhub/jupyterhub/issues/1244

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -389,20 +389,34 @@ class ConfigurableProxy extends EventEmitter {
   _handleProxyErrorDefault(code, kind, req, res) {
     // called when no custom error handler is registered,
     // or is registered and doesn't work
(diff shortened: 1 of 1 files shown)


parente:
Patch looks good to me. Happy to see this log cleanup.

willingc:
Thanks for the review @parente. LGTM too. Ran locally as well.

Thanks @minrk.

PR #126, opened 2017-09-18
release 3.0

minrk:
only minor intended changes, but major revision due to the refactoring to ES6.

cc @willingc


minrk:
Released 3.0

willingc:
Thanks @minrk :tada:

PR #129, opened 2017-10-06
Add a pluggable strategy for storage

jonathansp:
We'd like to suggest an alternative approach to solve https://github.com/jupyterhub/configurable-http-proxy/pull/85. This PR adds the ability to plug an external strategy for *configurable-http-proxy*. 

Currently, we have been working on an extension for redis: https://github.com/globocom/configurable-http-proxy-redis-backend

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -73,7 +73,8 @@ args
     "--proxy-timeout <n>",
     "Timeout (in millis) when proxy receives no response from target.",
     parseInt
-  );
+  )
+  .option("--storage-backend <storage-class>", "Define an external storage class. Defaults to in-MemoryStore.");
 
 args.parse(process.argv);
 
@@ -222,6 +223,9 @@ if (args.protocolRewrite) {
 if (!options.authToken) {
   log.warn("REST API is not authenticated.");
 }
+
+// external backend class
+options.storageBackend = args.storageBackend;
 
 var proxy = new ConfigurableProxy(options);
 
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -17,8 +17,7 @@ var http = require("http"),
   log = require("winston"),
   util = require("util"),
   URL = require("url"),
-  querystring = require("querystring"),
-  store = require("./store.js");
+  querystring = require("querystring");
 
 function bound(that, method) {
   // bind a method, to ensure `this=that` when it is called
@@ -129,13 +128,24 @@ function camelCaseify(options) {
   return camelOptions;
 }
 
+const loadStorage = (options) => {
+  if (options.storageBackend) {
+    const BackendStorageClass = require(options.storageBackend);
+    return new BackendStorageClass(options);
+  }
+
+  // loads default storage strategy
+  const store = require("./store.js");
+  return new store.MemoryStore(options);
+};
+
 class ConfigurableProxy extends EventEmitter {
   constructor(options) {
     super();
     var that = this;
     this.options = camelCaseify(options || {});
 
-    this._routes = new store.MemoryStore();
+    this._routes = loadStorage(options || {});
     this.authToken = this.options.authToken;
     if (options.includePrefix !== undefined) {
       this.includePrefix = options.includePrefix;
--- a/test/dummy-store.js
+++ b/test/dummy-store.js
@@ -1,0 +1,12 @@
+"use strict";
+
+class PlugableDummyStore {
+  get(path) {}
+  getTarget(path) {}
+  getAll() {}
+  add(path, data) {}
+  update(path, data) {}
+  remove(path) {}
+}
+
+module.exports = PlugableDummyStore;
--- a/test/proxy_spec.js
+++ b/test/proxy_spec.js
@@ -210,6 +210,27 @@ describe("Proxy Tests", function() {
     });
   });
 
+  it("options.storageBackend", function(done) {
+    const options = {
+      storageBackend: "mybackend",
+    };
+    expect(() => {
+      const cp = new ConfigurableProxy(options);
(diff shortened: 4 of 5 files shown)


yuvipanda:
Is this for use with JupyterHub? Or for other use cases?

yuvipanda:
(also, yay for your first PR to this repo! welcome :D)

jonathansp:
@yuvipanda we gonna use with jupyter hub, although it can be used in any case.

yuvipanda:
Personally, for JupyterHub I'd highly recommend implementing a custom proxy backend instead (http://jupyterhub.readthedocs.io/en/latest/reference/proxy.html) and using it with a different proxy (such as traefik) rather than adding more complexity to CHP. There's a half complete one for traefik + etcd in https://github.com/yuvipanda/jupyterhub-traefik - and more ones are easy to make. IMO that's the way forward rather than building and maintaining our own complex proxy implementation.

andrelu:
I think it would be nice to have a pluggable storage options. I think it does not add complexity to this project because the implementation would be outside.

jonathansp:
@yuvipanda implementing a custom proxy was our first thought. However, as discussed in #85, and it seems to be a recurring debate ( @minrk @rgbkrk @willingc @narenst ),  CHP is a great software and would be nice to have more backends. By following simple steps, we might build a redis-backend, a postgres-backend, an NFS backend etc.

lib/store.js
```diff
@@ -78,4 +78,4 @@ class MemoryStore extends BaseStore {
   }
 }
 
-exports.MemoryStore = MemoryStore;
+module.exports = MemoryStore;
```
minrk: I think this should be:

```javascript
exports.BaseStore = BaseStore;
exports.MemoryStore = MemoryStore;
```

so that the BaseStore is exported and can be used.

minrk:
This looks great! I'm very happy that you found a way to load the storage backend as a plugin without requiring it to be added to CHP itself. I left a minor comment inline. I imagine that at least some custom providers would want to `extend BaseStore`, so making sure that's exported is probably important.

Re: custom proxies, JupyterHub 0.8 introduces a Python API for its proxy requirements, making it more feasible to provide alternate implementations, as the Hub's API requirements are more clear.

lib/store.js
jonathansp: @minrk, by doing this I also need to move `BaseStore` to another file. The change I made expects a class to be exported, not a module. I did it because we don't need to know the class name we are importing (RedisStore, PGStore etc). Take a look:

```
const loadStorage = (options) => {
  const BackendStorageClass = require(options.storageBackend || "./store.js");
  return new BackendStorageClass(options);
};
```

Also, I'm not sure it is a good an idea for a plugin to use `BaseStore`. We might have cyclic references. What do you think?

minrk: Duck typing is fine, but BaseStore exists for the purpose of being used in different implementations, potentially to reduce how much has to be duplicated. In particular, it implements some basic and second-order functionality that reduces what implementations must provide (e.g. providers don't strictly need to implement `get`, only `getAll`). It also gives us a place to add new second-order methods that providers would get for free.

If we aren't exporting BaseStore, there's no reason for BaseStore to exist and it can be subsumed into MemoryStore as a single class, rather than the sole subclass of an unusable base.

We can split the files if we really have to, but maybe it would be better to allow it to be an attribute rather than requiring all providers to use a dedicated module that only exports one class? Perhaps something like `backend = './store.js:MemoryStore'`:

```javascript
[modName, className] = backend.split(':');
backendMod = require(modName);
if (className) {
  backendClass = backendMod[className];
else {
  backendClass = backendMod;
}
```

jonathansp: @minrk since it is not a javascript pattern (I suppose), I don't think it is a good idea. Unless you want to extend your storage strategies *inside*  the library, I vote for removing BaseStore. It's going to leave the code simpler.

rgbkrk: I think the main issue here is that this is going to cause a major bump in semver when we could easily make this a minor bump adding the functionality. Keep the original keys on the exported interface and it will be alright.

rgbkrk: Oh wait, I see that it's `exports` either way.

jonathansp: It exports only MemoryStore https://github.com/jupyterhub/configurable-http-proxy/blob/master/lib/store.js#L81

rgbkrk: Ok, my only suggestion then is to keep the original MemoryStore export, we can defer the `BaseStore` to outside the scope of this PR to at least finish bringing in the great work you've already done.

I agree with you about the potential for cyclic issues and for why the base store would need to end up in a separate file. I also see the ease at which a default export would provide. For now, let's keep it backwards compatible while moving forward.

jonathansp: @rgbkrk @minrk I made the changes. Could you please check again?

jonathansp:
Before merging this PR, let me add one more test case. I'll work on it tomorrow morning.

jonathansp:
Done: https://github.com/jonathansp/configurable-http-proxy/blob/master/test/proxy_spec.js#L223

minrk:
Cool!

willingc:
Thanks @jonathansp for the contribution 👍

jonathansp:
hi everybody, we would like to let you know we finished our [configurable-proxy-redis-backend](https://github.com/globocom/configurable-http-proxy-redis-backend/blob/master/lib/index.js). Please, feel free to make any suggestion!

jonathansp:
@willingc @minrk have you an idea when it's going to be released?

willingc:
@jonathansp I return from vacation tomorrow. I’ll make sure to touch base and see what else, if anything, @minrk would like in the release. Thanks again for the contributions.

minrk:
I think we can release 3.1 straight away with this (#131)

PR #131, opened 2017-11-02
changelog for 3.1

minrk:
will have support for custom backends

cc @willingc @jonathansp


willingc:
Thanks @minrk. Looks good :-)

PR #141, opened 2017-12-26
Fix: avoid calling updateLastActivity for each chunk of data

jonathansp:
Instead of waiting for the complete data, `updateLastActivity` is being called for every chunk of data. It's a wrong behavior since we should expect for `end` request's event. Registering this `data` callback hangs the application.

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -510,14 +510,16 @@ class ConfigurableProxy extends EventEmitter {
         that.handleProxyError(503, kind, req, res, e);
       });
 
-      // update timestamp on any reply data as well (this includes websocket data)
-      req.on("data", function() {
-        that.updateLastActivity(prefix);
-      });
+      if (kind === "ws") {
+        // update timestamp on any reply data as well (this includes websocket data)
+        req.on("data", function() {
+          that.updateLastActivity(prefix);
+        });
 
-      res.on("data", function() {
-        that.updateLastActivity(prefix);
-      });
+        res.on("data", function() {
+          that.updateLastActivity(prefix);
+        });
+      }
 
       // update last activity timestamp in routing table
       return that.updateLastActivity(prefix).then(function() {


rgbkrk:
Should update last activity be called on error too in case it terminates erroneously?

jonathansp:
Hi @rgbkrk , not sure, @willingc and @minrk could help us on it. If we want to update last activity in case of error, we can add a callback to listen to event `abort`.  Seems to be a simple change. I tried to put on this PR contains the smallest change so as to fix the problem, but I'm able to add this another callback if we decided to do.

rgbkrk:
All good, mostly just curiosity here as I haven't touched the proxy in a while.

minrk:
Registering activity on the data event is the correct behavior. This is because any data flowing through a kept-open connection (e.g. a websocket) is activity, which is the primary use case for CHP. Do you have an example test case where this causes a hang?

minrk:
Another option is to change activity tracking for persistent connections and instead refcount active connections. Instead of tracking every data chunk, return `now` for activity if any current connections are proxied for a given prefix. That's more complicated than tracking data events, but might be more performant.

If having the event listener is the problem, perhaps the more direct fix is to unregister the `data` event listener on `end` or `error`.

jonathansp:
Hi @minrk, we made a few tests here. It seems like by having a callback for `data` event causes a hang when used along with our `redis-backend`. According to the docs, `data` callback is triggered when body data is sent, for example in POST or PUT. For websocket connections, `data` callback is used to handle every message from a client. After debugging, we realized it freezes only when function  `handleProxy(kind, req, res)` is called as `kind` being `web`. It does not happen for websocket.

Also, for POST and PUT, `updateLastActivity` have been called twice per request.  Once here ([line 514](https://github.com/jupyterhub/configurable-http-proxy/blob/master/lib/configproxy.js#L514)):
```javascript
        req.on("data", function() {
          that.updateLastActivity(prefix);
        });
```
and when the function returns ([line 525](https://github.com/jupyterhub/configurable-http-proxy/blob/master/lib/configproxy.js#L525)):

```javascript
      return that.updateLastActivity(prefix).then(function() {
        // dispatch the actual method
        that.proxy[kind].apply(that.proxy, args);
      });
```

As you commented, `data` callback is important to register WebSocket connection activities, so I changed the PR in order to keep this behavior. However, I kept the callback only for websocket, since for regular http connections, `updateLastActivity` works on line 525.

jonathansp:
Steps to reproduce the problem:

1. Install _redis_, _jupyterhub_, _configurable-http-proxy_ and _configurable-http-proxy-redis-backend_.
2. Run `CONFIGURABLE_PROXY_REDIS_URI=redis://localhost jupyterhub --JupyterHub.proxy_cmd="['configurable-http-proxy', '--storage-backend', 'configurable-http-proxy-redis-backend']"`
3. Open your browser, create an empty notebook and open it.
4. Kernel won't be load because a POST to http://localhost:8000/user/jonathanprates/api/sessions will never ends.

Note: I'm not dismissing a problem in _configurable-http-proxy-redis-backend_ eventually but I haven't found any there.

minrk:
Since the same thing works fine without the redis backend and not with the redis backend, it seems like there's something amiss in the backend. That doesn't mean there's nothing we should change, necessarily, since it could be a timing issue that doesn't come up for the in-memory case. We should also still be tracking data on non-websocket messages for things like event streams or other longstanding requests, so making it conditional on websockets isn't sufficient.

Maybe there's an issue in the redis backend that's throwing an error that isn't getting logged?

jonathansp:
@minrk were you able to reproduce the problem?

> We should also still be tracking data on non-websocket messages for things like event streams or other longstanding requests, so making it conditional on websockets isn't sufficient.

As I said, for non-websockets messages, it already happens [here](https://github.com/jupyterhub/configurable-http-proxy/blob/master/lib/configproxy.js#L523).

minrk:
> As I said, for non-websockets messages, it already happens here.

That's only the start of the request, not the intermediate 'messages' on things like an event stream. It must happen on the `data` event (or some other event to continuously update), because a single request should be recording activity not just at the start and end, which may be several minutes or more apart, but throughout. Alternately, we would need to have bookkeeping to ensure that active connections are accounted for.

> were you able to reproduce the problem?

Yes, with redis. Not with the existing backend. That suggests that redis is related to the problem, either via a bug in the backend or a sensitivity to timing in CHP, where the additional time of calling out to redis is causing the state to be different.

minrk:
Worked it out. The promise that's been added in between activity monitoring and hooking up the activity updates to the data event allowed the request body to be consumed before the proxy started forwarding it. Ensuring there are no async events in between hooking up the proxy and the activity monitoring fixes the problem without any other changes.

jonathansp:
Sounds great! Thanks, @minrk !

minrk:
Will do a bugfix release for this when I get the chance, possibly around the end of the week, depending on time.

minrk:
Released 3.1.1 with this fix

PR #163, opened 2018-05-25
[wip] ensure optionally-decoded characters (@, ~) are treated the same as their encoded counterparts

minrk:
we already had it working one direction (%40->@), but not the other (@->%40)

related to jupyterhub/jupyterhub#1717 but not actually necessary for the fix

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -370,7 +370,7 @@ class ConfigurableProxy extends EventEmitter {
     var timer = this.statsd.createTimer("find_target_for_req");
     // return proxy target for a given url path
     var basePath = this.hostRouting ? "/" + parseHost(req) : "";
-    var path = basePath + decodeURIComponent(URL.parse(req.url).pathname);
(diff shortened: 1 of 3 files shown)


minrk:
marking WIP since there seems to be something wonky

Pugio:
Is there something blocking this? I have an issue with `%40` in usernames (`/user/myname@mydomain/` vs `/user/myname%40mydomain/`) returning 404s, which I think this PR would fix.

While my issue is with using `code-server` behind the jupyter-server-proxy, it is functionally identical to the problem discussed in jupyterhub/jupyterhub#2547.

manics:
Please see the discussion on https://github.com/jupyterhub/jupyterhub/pull/3167. This is tricky to get right.

MTTrackNotion77 (approved):

yuvipanda:
Given that @minrk has closed https://github.com/jupyterhub/jupyterhub/pull/3167, I'm going to close this PR as well. Feel free to re-open if you disagree, @minrk.

PR #170, opened 2018-07-23
End to end ssl

tgmachina:
These changes are in service of turning on SSL internally for JupyterHub in reference to [JupyterHub issue 371](https://github.com/jupyterhub/jupyterhub/issues/370).

The change amounts to creating a URL object from a stored target url string. This allows for attaching certs to a request to a target.

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -248,22 +248,8 @@ class ConfigurableProxy extends EventEmitter {
 
     var that = this;
 
-    // Parse target into object to attach ssl params. The underlying
-    // http-proxy library will use this to attach ssl to the forwarded
-    // request.
-    data.target = URL.parse(data.target);
-    if(this.options.api_ssl){
-        data.target.key = this.options.api_ssl.key;
-        data.target.cert = this.options.api_ssl.cert;
-        data.target.ca = this.options.api_ssl.ca;
-    }
-
-    this._routes.add(path, data, function () {
-      that.update_last_activity(path, function () {
-        if (typeof(cb) === "function") {
-          cb();
-        }
-      });
+    return this._routes.add(path, data).then(() => {
+      this.updateLastActivity(path);
     });
   }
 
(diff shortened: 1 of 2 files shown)


lib/configproxy.js
```diff
@@ -501,6 +502,13 @@ class ConfigurableProxy extends EventEmitter {
       if (!that.includePrefix) {
         req.url = req.url.slice(prefix.length);
       }
+      target = URL.parse(target);
+      if(that.options.apiSsl){
+          target.key = that.options.apiSsl.key;
+          target.cert = that.options.apiSsl.cert;
+          target.ca = that.options.apiSsl.ca;
```
minrk: Is it right to use the same SSL key and certificate that the proxy is using for requests *to* the API (apiSsl) and requests from the proxy to proxy targets?

Essentially, I'm wondering if we need a separate client ssl certificate/config for these requests.

tgmachina: Good question! I also debated this myself. The mental model I've been using is that CHP has a both a front and back-end identity represented by `ssl` certs `apiSsl` certs respectively.

I do like having the option to specify a different set of certs (creating a second back-end identity). Originally, I had a `--forward-ssl` flag where I had envisioned attaching a separate set of certs and options in much the same way `--api-ssl` is used. Since CHP is not exclusively used with JupyterHub, I can see having this option as being useful to satisfy some use-case I haven't thought of.

The potential concerns--that I see anyway--are that in using the same certs, we inadvertently allow a client of the API to establish ssl trust directly with a target of the API or that a client could impersonate a target and vice versa. The former is avoidable by using separate CAs for each entity, but the latter would not be avoidable at present. 

In short, having a separate set of certs going out to targets would probably be for the best overall. Objections to my former naming of `--forward-ssl`? Maybe `--target-ssl`? Naming things is hard...

minrk: I would call it `--client-ssl` and support only a single ssl client certificate for the proxy, for all outgoing SSL connections, and see if we can get away with that.

The alternative I see is to add to the *API* the ability to add SSL client info when adding a route, e.g. it comes in on a POST request, separate for each connection.

PR #186, opened 2018-12-21
Update the instruction for required Node.js version

dtaniwaki:
This package requires Node.js v6 [instead of v4 now](https://github.com/jupyterhub/configurable-http-proxy/commit/f824c65d9bf7d8a4d050b9fbfd8db65dcde65f6d)

I updated the README so people don't get confused and spend quite a few time to debug syntax errors such as destructuring.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,9 @@ functionality to [JupyterHub] deployments.
 
 ## Install
 
-Prerequisite: [Node.js](https://nodejs.org/en/download/) ≥ 4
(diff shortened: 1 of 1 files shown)


dtaniwaki:
Actually, just installing by the nodesource script doesn't work because it creates `nodejs` and no `node` alias. We still need to make an alias or change [the shebang](https://github.com/jupyterhub/configurable-http-proxy/blob/e4eef778de07ccbdd10295c7424985b6228979c6/bin/configurable-http-proxy#L1)

Issue #120, opened 2017-08-17
Can not specify which port / address needs to be used when redirecting http -> https


yuvipanda:
Related to https://github.com/jupyterhub/jupyterhub/issues/1289.

Use case:

CHP is listening on port 8000 (since it's running as user Nobody) but kubernetes maps that to port 80 on an external IP. We want to do HTTP -> HTTPS redirects here, but CHP assumes that bind IP == connect IP, and hence redirects users to its local IP address, which can not be resolved from the external internet.

minrk:
I don't think this is related to #1289 because unlike the Hub, CHP never specifies from internal configuration what ip anyone should connect to.

CHP doesn't use the bind address (which is ~always 0.0.0.0) for this, it uses the Host header so that it should be the same as the connect ip for a properly assembled request. My guess is the upstream proxy is not passing through the Host header. We could try to figure out what the proxied Host was (e.g. `X-Forwarded-Host`), or we could require that the upstream proxy [preserve host headers](https://docs.oracle.com/cd/E40518_01/studio.310/studio_install/src/cidi_studio_reverse_proxy_preserve_host_headers.html) for this feature.

Can you inspect the headers to see what kubernetes is passing?

If there is an additional upstream proxy, it probably makes more sense for *it* to do the HTTP->HTTPS redirect, anyway.

clancychilds:
I think, IIUC, the problem is less about the host, and more about the port?
Inside the k8s cluster, the HTTPS service is listening on 8443 and the HTTP service on 8000. These are portforwarded on the loadbalancer from 443 and 80 respectively.
When `--port 8443` and `--redirect-port 8000` are set, then a user trying to connect to <external>:80 is portforwarded to <internal>:8000 which is 301'd to <external>:8443, which doesn't work.
It may need an option like `--redirect-to 443` which would be used on https://github.com/jupyterhub/configurable-http-proxy/blob/master/bin/configurable-http-proxy#L349 to that port (rather than the listen.port)

minrk:
closed by #187

PR #187, opened 2019-01-03
Command line argument --redirect-to adds the ability to specify the port to forward HTTP

clancychilds:
This addresses the problem that we have in https://github.com/jupyterhub/configurable-http-proxy/issues/120

If, for instance, a loadbalancer or other NAT has 443 redirected to 8443 and CHP thinks the port is 8443, it would redirect HTTP traffic to that port, which might not work. This allows `--redirect-to 443` as an option on CHP to fix this.

I also added some tests of the CLI app, as this functionality lives inside there and was otherwise missing tests.

(When this is accepted and published, I will work to fix the jupyterhub helm chart for kubernetes to use this option.)


minrk:
Makes sense, thanks!

clancychilds:
@minrk I didn't bump the version number. Was I supposed to? Is there a way to get this tagged with a new version in docker hub (i.e. 3.1.2) so that it can be specified from a docker hub pull?

PR #189, opened 2019-01-09
corrected request body for Add proxy request

rushikeshraut777:
This Pr contains correction in the README file.

The request body for create proxy endpoint POST /api/route/[:path] is corrected.


minrk:
Thanks!

PR #206, opened 2019-06-03
Command line option for custom headers

### LINKED ISSUE #205: Command line option for custom headers
`configurable-http-proxy` supports a wide range of configuration options, but it is not practical for it to cover all use cases. An example of an uncovered use case is the [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)  response header. Instead of adding yet another configuration option specifically for this header I propose we implement a `--custom-headers <headers>` configuration option that allows for user-specified headers to be added to all proxied requests. Upstream projects like https://github.com/jupyterhub/zero-to-jupyterhub-k8s can expose the option to be passed to `configurable-http-proxy`, for example through the Helm chart.

ivan-gomes:
Resolves #205.

> `configurable-http-proxy` supports a wide range of configuration options, but it is not practical for it to cover all use cases. An example of an uncovered use case is the [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)  response header. Instead of adding yet another configuration option specifically for this header I propose we implement a `--custom-headers <headers>` configuration option that allows for user-specified headers to be added to all proxied requests. Upstream projects like https://github.com/jupyterhub/zero-to-jupyterhub-k8s can expose the option to be passed to `configurable-http-proxy`, for example through the Helm chart.

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -82,8 +82,9 @@ args
     "Rewrite the Location header protocol in redirect responses to the specified protocol"
   )
   .option(
-    "--custom-headers <headers>",
-    "Comma separated list of custom headers to add to proxied requests (k1:v1,k2:v2,...)"
+    "--custom-header <header>",
+    "Custom header to add to proxied requests. Use same option for multiple headers (--custom-header k1:v1 --custom-header k2:v2)",
+    collect, []
   )
   .option("--insecure", "Disable SSL cert verification")
   .option("--host-routing", "Use host routing (host as first level of path)")
@@ -100,6 +101,10 @@ args
     "--storage-backend <storage-class>",
     "Define an external storage class. Defaults to in-MemoryStore."
   );
+
+function collect(value, previous) {
+  return previous.concat([value]);
+}
 
 args.parse(process.argv);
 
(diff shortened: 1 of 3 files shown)


rgbkrk:
My only main question is whether these headers are applied after other headers being set and if we should document so here.

ivan-gomes:
Custom headers are indeed applied after all other headers using the `proxyRes` callback.

bin/configurable-http-proxy
```diff
@@ -294,6 +298,24 @@ if (!options.authToken) {
   log.warn("REST API is not authenticated.");
 }
 
+// custom headers option
+options.customHeaders = [];
+if (args.customHeaders) {
+  options.customHeaders = args.customHeaders.toString().split(',').map(s => s.trim()).map(header => {
```
minrk: I don't think simple splitting on commas is going to work, since commas could be in the header values. Maybe try following the [collect example](https://github.com/tj/commander.js/#custom-option-processing) for specifying multiple values for an option, so it would look like:

```
chp --header "a: x,y,z" --header "b: a,b,c"
```

bin/configurable-http-proxy
```diff
@@ -294,6 +298,24 @@ if (!options.authToken) {
...
+    var i = header.indexOf(':');
+    var key, value;
+    if (i < 0) {
+      key = header;
```
minrk: Probably error here would be more useful than assuming an empty string for invalid input?

rgbkrk:
Min's points are solid so I've set myself to pending review (though GitHub still shows my approval).

bin/configurable-http-proxy
ivan-gomes: Implemented.

bin/configurable-http-proxy
ivan-gomes: Implemented.

ivan-gomes:
Implemented @minrk's points. Thanks for the guidance. @rgbkrk would you mind re-approving?

minrk:
Thanks! Sorry for being so slow

PR #209, opened 2019-07-24
Add file_env function to set the token env var

rcthomas:
This enables setting `CONFIGPROXY_AUTH_TOKEN` using e.g. Docker or Rancher secrets.


minrk:
Nice!

PR #211, opened 2019-10-01
Change user to numeric value for k8s compatibility

m2hofi94:
When a k8s cluster enforces the PodSecurityPolicy "MustRunAsNonRoot" it requires a numeric user to verify, as per documentation https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups.

I therefore simply changed the `USER` to the numeric equivalent of `nobody`.


minrk:
Thanks!

PR #213, opened 2019-11-08
chore: Udpate node, replace add with copy

jgwerner:
- Update Node to v12.13
- Use COPY instead of ADD


minrk:
Thanks!

consideRatio:
After this merged PR, port 8000 is no longer exposed by default when creating a docker container, is that an issue?

PR #216, opened 2019-11-11
Fix Vulnerabilities

rafael-ladislau:
The packages "winston" and "loadsh" was getting listed as vulnerable.
After the command "npm audit fix", the project is using not vulnerable
packages.
And the "npm test" command is running without any failures.

We are running JupyterHub in a FedRamp compliant environment, and We've checked the vulnerability after our regular security scans.

It would be great if we could use the Docker image you provide with this fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -194,9 +194,9 @@
       },
       "dependencies": {
         "lodash": {
-          "version": "4.17.11",
-          "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.11.tgz",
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks! LGTM!

rafael-ladislau:
Last question, do you have any idea when you intend to release a new Docker Image with this fix? It would be able if we could use the image provided by you with this fix in our environment.

minrk:
@rafael-ladislau you can expect 4.2.0 today

consideRatio:
4.2.0 is out!

PR #220, opened 2019-11-12
RELEASE.md documentation and small fixes

consideRatio:
## PR Summary
- Added a dockerhub badge to the readme
- Started testing Node version 12 in travis
- Exposed port 8000 and 8001 in the Dockerfile, which doesn't mean you automatically have them exposed if you run `docker run` but indicates two ports in use by the software within the Dockerfile that could very well want to be _published_ by the `-P` flag or `-p 8000:8000` flag for example.
- Added my best attempted documentation on how to make a release in a RELEASE.md file. This comes to a large extent from a Chartpress PR (https://github.com/jupyterhub/chartpress/pull/74) by @minrk. I adopted bump2version as I saw a .bumpversion.cfg file in this repo.
  - Changelog
  - Version bumping and git commit tagging
  - Pushing to NPM
  - Resetting the version to `-dev`

## Help requested (RESOLVED)
I note that the [build is failing on Node 12](https://travis-ci.org/jupyterhub/configurable-http-proxy/jobs/611011264?utm_medium=notification&utm_source=github_status), yet I know that we bumped the [Dockerfile](https://github.com/jupyterhub/configurable-http-proxy/blob/master/Dockerfile) to start with a `FROM node:12.13-alpine` statement recently in #213. Solving this is very relevant I figure.

**UPDATE:** I fixed this issue in 7238f2b by relaxing the test a bit... It may be caused by an upstream bug in node or jasmine but I'm not in ready to search and fix that at the moment. The test failure said that it expected "asdf" but got "asdf" more or less, which is very weird to me. I assumed there could be some type difference etc so I started looking for if the error message contained the relevant string instead of explicitly testing for a matching error object.

## Review suggestions (RESOLVED)
I stopped testing of node 6 on a gut feeling when adding node 12, should I keep support for node 6 perhaps? Should we drop testing of node 6 and also adjust package.json's constraint on the node version to be node >= 8?

**RESOLVED:** We continue to test and allow for node 6 to be used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ branches:
   except:
     - /^greenkeeper.*/
 node_js:
+  - 12
   - 10
   - 8
   - 6
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM node:12.13-alpine
+# ref: https://hub.docker.com/_/node?tab=tags&name=12
 
 LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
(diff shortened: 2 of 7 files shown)


minrk:
This is fine. There's no real need for us to drop node 6 support, though nodejs itself has dropped it. If you wanted to revert that change, this is fine with me. We can save that for a major release.

consideRatio:
Nice okay @minrk!
- I removed the steps about building and pushing the Docker image and instead added a note about this and linked to the dockerhub build status.
- I reverted all changes relating to dropping node 6 support, both testing and requiring node 8 or above.

PR #227, opened 2020-03-11
Merge pull request #227 from consideRatio/production-dockerfile

consideRatio:
Running npm install on a local folder will by default also install the
devDependencies specified in package.json, and this is not suitable when
we are building a Dockerfile for use in production.


minrk:
Makes sense!

PR #228, opened 2020-03-12
move `npm audit` to CI from Dockerfile

minrk:
If this new test stage fails, running `npm audit fix` and committing the result will fix it.

Moves the fixes to the repo's package-lock.json, rather than deferring to the docker build. This means that all installs of CHP will get fixes, not just docker, and ensures that package-lock.json accurately represents what is installed in the image.

It also ensures that we are notified when a fix is needed.

The base image is also pinned only to the major node version to ensure that security patches are received on every build, rather than forcing us to keep bumping the base image.

This also reverts the removal of npm in #226, which didn't have any effect on the vulnerability of the image, but there may be some unfortunate policy reasons why it needs to be done, despite having no effect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ install:
   - npm install -g codecov
   - npm install
 script:
+  - npm audit # if npm audit fails, run `npm audit fix` and commit the results
   - npm run lint
   - travis_retry npm test
 after_success:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.16.1-alpine
+FROM node:12-alpine
 # ref: https://hub.docker.com/_/node?tab=tags&name=12
 
 LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
@@ -10,11 +10,8 @@ RUN mkdir -p /srv/configurable-http-proxy
 COPY . /srv/configurable-http-proxy
 WORKDIR /srv/configurable-http-proxy
 
-# Install configurable-http-proxy, then automatically install compatible updates
-# to vulnerable dependencies, and finally uninstall npm which isn't needed.
-RUN npm install -g --production \
- && npm audit fix \
- && npm uninstall -g npm
+# Install configurable-http-proxy
+RUN npm install -g --production
 
 # Switch from the root user to the nobody user
 USER 65534


.travis.yml
```diff
@@ -12,6 +12,7 @@ install:
   - npm install -g codecov
   - npm install
 script:
+  - npm audit # if npm audit fails, run `npm audit fix` and commit the results
```
consideRatio: I did `npm audit --help` to try figure out what is audited.

```
npm audit [--json] [--production]
npm audit fix [--force|--package-lock-only|--dry-run|--production|--only=(dev|prod)]
```

Perhaps it make sense to use `--production` in the `npm audit` to not have it fail on devDependencies be considered. 

## Rubber duck...

In [the documentation for npm audit](https://docs.npmjs.com/cli/audit) I learned that it is scrubbed content of package-lock.json that is submitted for review as compared to the package.json for example. So, if we would do `npm audit` we would realize if package-lock.json could use an update, and if we would do `npm audit fix` we would update the package-lock.json file... And by doing `npm audit fix --force` we would also potentially update the `package.json` file as I understand it.

Okay I think I've concluded now that I agree that we should do this in travis, and ensure our git repo is updated with a seemingly secure package-lock.json at all time. I also now hypothesize that `npm audit fix` to only update package-lock.json given the constraints of `package.json`, while `npm audit fix --force` will also update package.json.

## Questions
- Should we use `--production` in this audit to not bug with script failures? I think yes.
- Should we suggest the `npm audit fix` to be run with `--only=prod` to avoid bothering with dev dependencies? I think we should update it all if we update the prod environment.

consideRatio:
```suggestion
  - npm audit --production # if npm audit fails, run `npm audit fix` and commit the results
```

consideRatio:
> avoid calling `npm audit fix` in the image, which results in package-lock.json not being representative of what's in the image

If we want the docker image to represent our in-repo package-lock.json, we should ensure we install from package-lock.json, which I think we currently don't when we write `npm install --production` that will update package-lock.json.

I think then perhaps what we need is `npm ci`, see: https://docs.npmjs.com/cli/ci.html#description

.travis.yml
```diff
@@ -12,6 +12,7 @@ install:
   - npm install -g codecov
   - npm install
```
consideRatio:
```suggestion
  - npm ci
```

consideRatio: Discussion in https://github.com/jupyterhub/configurable-http-proxy/pull/228#issuecomment-598153327

Dockerfile
```diff
@@ -10,11 +10,8 @@ RUN mkdir -p /srv/configurable-http-proxy
 COPY . /srv/configurable-http-proxy
 WORKDIR /srv/configurable-http-proxy
 
-# Install configurable-http-proxy, then automatically install compatible updates
-# to vulnerable dependencies, and finally uninstall npm which isn't needed.
-RUN npm install -g --production \
- && npm audit fix \
- && npm uninstall -g npm
+# Install configurable-http-proxy
+RUN npm install -g --production
```
consideRatio:
```suggestion
RUN npm ci --production
```

I wanted us to start using `npm ci` to work against package-lock.json, but we can not use `-g` with `npm ci`, so that need to be resolved somehow as the `chp-docker-entrypoint` script assumes `configurable-http-proxy` to be on path.

PR #239, opened 2020-05-29
CI: npm-audit cronjob in travis

consideRatio:
A PR relating to @minrk's #228.

We add a TravisCI job definition, that will only run on TravisCI configured CronJobs, and then we configure this to run daily, and we get informed about the need to do run `npm audit fix`.


consideRatio:
Merging this to get this tested properly, which requires it to be part of master.

PR #241, opened 2020-05-29
Docker image: use package-lock.json and only include relevant parts

consideRatio:
Together with the TravisCI Job definition added in #239 / #240 and [configured to run in TravisCI settings](https://travis-ci.org/github/jupyterhub/configurable-http-proxy/settings), I suggest this PR replace #228.

Closes #228.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+# Ignore everything...
+*
+# ... except:
+!bin/
+!lib/
+!chp-docker-entrypoint
+!index.js
+!package*.json
+!LICENSE
--- a/chp-docker-entrypoint
+++ b/chp-docker-entrypoint
@@ -41,4 +41,4 @@ case "$@" in
     ARGS="--api-ip=0.0.0.0";;
 esac
 
-exec bin/configurable-http-proxy $ARGS "$@"
+exec configurable-http-proxy $ARGS "$@"
--- a/chp-docker-entrypoint
+++ b/chp-docker-entrypoint
@@ -41,4 +41,4 @@ case "$@" in
     ARGS="--api-ip=0.0.0.0";;
 esac
 
-exec configurable-http-proxy $ARGS "$@"
+exec bin/configurable-http-proxy $ARGS "$@"
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apk add --no-cache jq curl
 RUN mkdir -p /srv/configurable-http-proxy
 COPY . /srv/configurable-http-proxy/
 WORKDIR /srv/configurable-http-proxy
+ENV PATH=/srv/configurable-http-proxy/bin:$PATH
 
 # Install configurable-http-proxy according to package-lock.json (ci) without
 # devDepdendencies (--production), then uninstall npm which isn't needed.
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,10 @@ LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 # Useful tools for debugging
 RUN apk add --no-cache jq curl
 
+# Copy relevant (see .dockerignore)
 RUN mkdir -p /srv/configurable-http-proxy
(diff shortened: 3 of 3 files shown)


chp-docker-entrypoint
```diff
@@ -41,4 +41,4 @@ case "$@" in
     ARGS="--api-ip=0.0.0.0";;
 esac
 
-exec configurable-http-proxy $ARGS "$@"
+exec bin/configurable-http-proxy $ARGS "$@"
```
consideRatio, line 44: ## Explanation of line
Due to the change from `npm install -g` to `npm ci` where the `-g` flag isn't available, we need to reference the local executable instead.

I tried copying this to /usr/local/bin but then I got the following error, so I think this is a good solution.

```
/srv/configurable-http-proxy $ ./chp-docker-entrypoint 
internal/modules/cjs/loader.js:969
  throw err;
  ^

Error: Cannot find module '../package.json'
Require stack:
- /usr/local/bin/configurable-http-proxy
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:15)
    at Function.Module._load (internal/modules/cjs/loader.js:842:27)
    at Module.require (internal/modules/cjs/loader.js:1026:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/usr/local/bin/configurable-http-proxy:11:9)
    at Module._compile (internal/modules/cjs/loader.js:1138:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/usr/local/bin/configurable-http-proxy' ]
}
```

.dockerignore
```diff
@@ -0,0 +1,9 @@
+# Ignore everything...
+*
+# ... except:
+!bin/
+!lib/
+!chp-docker-entrypoint
+!index.js
+!package*.json
+!LICENSE
```
consideRatio, line 9: I added this file because I noted that we copied things from node_packages and felt that it could risk causing some issue, and that it would trigger rebuilds for no use potentially.

I got a warning about the node_packages being copied and existing, this is what led me to this change. It will come with the drawback of needing to update this as well if we add something to include, but hopefully we will add that within the lib folder.

minrk: I think instead of setting what gets added in dockerignore, we should whitelist in COPY directives instead. What do you think?

consideRatio: Ah I did that first, and then switched to this option for aesthetics.

When I concluded it was required to look something like below, with three lines instead of one, I felt better about a `.dockerignore` and `COPY . /dest` than multiple `COPY ` statements - one for each folder and one of the files. It turned out that one cannot copy an entire folder next to files files in a COPY statement, then the contents of the folder is copied to the destination rather than the entire folder.

```Dockerfile
# the option to a .dockerignore file afaik
COPY our-bin-folder /destination
COPY our-lib-folder /destination
COPY list-of-individual-files /destination
```

Let me know what you think, then I'll go with that!

consideRatio: Ping @minrk

chp-docker-entrypoint
consideRatio: Hmmm while it works, because the current working directory is the same as this files location, but I think it is a bit unrobust. So, I'll add the folder to path instead from the Dockerfile.

consideRatio: I updated the Dockerfile's PATH env var to find the ./bin folder

.dockerignore
consideRatio: I added a comment about the .dockerignore file the risk of missing out on what goes on is minimized.

minrk:
Thanks! I think we can reach a better compromise in .dockerignore, but this is still an improvement.

PR #242, opened 2020-05-29
fix: --custom-header flag implementation

### LINKED ISSUE #223: The --custom-header doesn't do anything
I want to send a custom header to backend servers. I think the only way is with "--custom-header" argument, but if It isn't correct please help me.

So, when I run CHP with "--custom-header" I don't watch this be running correctly.

For example:

I run CHP with this options:

```
> node_modules/.bin/configurable-http-proxy --ip 0.0.0.0 --port 8080 --api-ip 127.0.0.1 --api-port 8081 --log-level debug --custom-header "Authorization":"token kYFkuyfKUYf"
```

Then the backend server should receive the header, but NO!

```
127.0.0.1 - - [20/Dec/2019 13:10:22] "GET / HTTP/1.1" 200 -
{'X-Forwarded-Host': 'localhost:8080', 'X-Forwarded-Proto': 'http', 'X-Forwarded-Port': '8080', 'X-Forwarded-For': '127.0.0.1', 'Cache-Control': 'no-cache', 'Pragma': 'no-cache', 'Upgrade-Insecure-Requests': '1', 'Cookie': 'io=gkj_1lr05tkDtKyqAAAJ; __wzd22063e6fcb4b21b39a2f=1574782783|3600d444d719; ajs_user_id=null; ajs_group_id=null', 'Connection': 'close', 'Accept-Encoding': 'gzip, deflate', 'Accept-Language': 'es,en;q=0.5', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8', 'Host': 'localhost:8080'}
```

What is happening?

consideRatio:
The `--custom-header` flag added in #206 didn't seem to work, this PR attempts to fix this and also adds a test to make sure.

**Given current node-http-server, the key fixes are:**
- A singular/plural typo for `customHeader(s)`
- node-http-proxy configuration API, it now needs to be [`headers` and not `customHeaders`](https://github.com/http-party/node-http-proxy#options)
- `headers` now needs to get an object, not a list of objects

Closes #223.

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -145,7 +145,7 @@ class ConfigurableProxy extends EventEmitter {
     } else {
       this.includePrefix = true;
     }
-    this.customHeaders = this.options.customHeaders;
+    this.headers = this.options.headers;
     this.hostRouting = this.options.hostRouting;
     this.errorTarget = options.errorTarget;
     if (this.errorTarget && this.errorTarget.slice(-1) !== "/") {
@@ -229,9 +229,6 @@ class ConfigurableProxy extends EventEmitter {
 
     this.proxy.on("proxyRes", function(proxyRes, req, res) {
       that.statsd.increment("requests." + proxyRes.statusCode, 1);
-      if (that.customHeaders) {
-        that.customHeaders.forEach(header => res.setHeader(header.key, header.value));
-      }
     });
   }
 
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -84,7 +84,7 @@ args
   .option(
     "--custom-header <header>",
     "Custom header to add to proxied requests. Use same option for multiple headers (--custom-header k1:v1 --custom-header k2:v2)",
-    collect, []
+    collectHeadersIntoObject, {}
   )
   .option("--insecure", "Disable SSL cert verification")
   .option("--host-routing", "Use host routing (host as first level of path)")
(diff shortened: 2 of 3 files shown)


lib/configproxy.js
```diff
@@ -145,7 +145,7 @@ class ConfigurableProxy extends EventEmitter {
     } else {
       this.includePrefix = true;
     }
-    this.customHeaders = this.options.customHeaders;
+    this.headers = this.options.headers;
```
minrk, line 148: the `customHeaders` name is still used on L232. Can you remove it?

consideRatio: Woops, nice catch. Note to self: always search for something I replace to spot if I missed something.

consideRatio: Hmmm... so this is a part of the constructor of ConfigurableProxy

```node
    // proxy websockets
    this.proxyServer.on("upgrade", bound(this, this.handleProxyWs));

    this.proxy.on("proxyRes", function(proxyRes, req, res) {
      that.statsd.increment("requests." + proxyRes.statusCode, 1);
      if (that.customHeaders) {
        that.customHeaders.forEach(header => res.setHeader(header.key, header.value));
      }
    });
```

If I replace this with `that.header` I get errors I get socket hang up errors and I'm not sure why. Debugging.

consideRatio: I'm not sure what this code is meant to do or what the "proxyRes" event is, is that a response event to a request made? Then it doesn't make sense to modify the response.

Hmm...

consideRatio: Also, httpProxy should be configured already.

I'll delete the section, perhaps that is what you meant with "can you remove it" rather than "can I replace the use of customHeaders with headers".

consideRatio: I deleted that section, which previously have done nothing, and seem to contain logic of no use.

consideRatio:
@minrk I think this is ready for merge

consideRatio:
As the current implementation isn't working, and I've verified this implementation works with tests. I'll go ahead and self-merge this tomorrow.

consideRatio:
I did a another self review of my code and became more confident the implementation to fix the current broken logic makes sense. I'll go ahead and self-merge this.

PR #248, opened 2020-07-04
Emit proxyRequestWs events correctly, and some inline docs

consideRatio:
It seems like we use the emit function to trigger potential event
handlers registered, but from what I understand we should do this
differently depending on either proxied web requests and proxied web
socket requests.


minrk:
Great!

PR #254, opened 2020-08-11
lib: Use client ssl config to access error target

chancez:
Fixes usage of self-signed certs with Jupyterhub with CHP.
Original issue here:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1742#issuecomment-671617614

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -421,7 +421,6 @@ class ConfigurableProxy extends EventEmitter {
     // custom error server gets `/CODE?url=/escapedUrl/`, e.g.
     // /404?url=%2Fuser%2Ffoo
 
-    var proxy = this;
     var errMsg = "";
     this.statsd.increment("requests." + code, 1);
     if (e) {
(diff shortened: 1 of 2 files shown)


chancez:
I'm not much of JS guy, so there's probably better ways to do some of this.

consideRatio:
@chancez could you describe the logic change made so i can have an easier time reviewing this PR? I'm not also great with this tech stack.

chancez:
First, error-target is the URL the CHP will request if there is an error processing a request, like if a route doesn't exist.

Basically, for everything besides the error target, the CHP was using the node-http-proxy proxy object to handle sending requests to the various backends for each route. However, for the error-target, it was using it's own logic. This logic used a default HTTP client, and basically took the headers and body of the response and returned those as the response for the current request.

This is problematic, because the HTTP client, using it's defaults, obviously cannot validate self-signed certificates. Instead of using it's our own HTTP client to make requests to the error target, it uses the proxy object from node-http-proxy to do the request, just like it does for any other routes. The proxy object is configured with the CAs to validate self-signed certificate, so it fixes the issue. In addition, it makes the handling of errors the same as handling other requests for these purposes. This simplifies the logic of returning the response as well, since it's not error prone to copying the data/headers from the upstream response.

consideRatio:
This sounds sensible to me @chancez, I fail to resolve the CI errors or understand this in depth enough to make this PR go through at this point. Should we request help?

chancez:
Yeah, I'm not sure. I've been running this in production for a few months, so anything to get it merged would be great.

minrk:
Thanks for your patience on this one!

PR #255, opened 2020-08-25
simplify dockerignore to exclude node_modules

minrk:
instead of explicitly including every path

this results in including a few small, not strictly necessary files, but will be much simpler to maintain


consideRatio:
Nice thanks Min!

PR #256, opened 2020-08-25
try dependabot for updates

minrk:
try switching from renovate to GitHub's own [dependabot](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "javascript"
```
manics, line 6: `javascript` isn't listed on https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem

It's listed on https://dependabot.com/docs/config-file/ but that's for version 1

minrk: gotcha, I got confused by the multiple versions

PR #290, opened 2021-03-04
Improvements when things go wrong

minrk:
Investigating improvements that could be related to the increase in CHP load on mybinder.org which is currently seeing an increase in routes to stopped endpoints. These are currently giving mostly EHOSTUNREACH, and unhandled `ERR_STREAM_WRITE_AFTER_END` tracebacks, specifically in [this request](https://github.com/jupyterhub/configurable-http-proxy/blob/6792cebc5e2a8d28c258fbd652da1a9370472fa3/lib/configproxy.js#L472) to the upstream error-target provider (jupyterhub).

- avoid write-after-end by checking [res.writableEnded](https://nodejs.org/api/http.html#http_response_writableended) (thanks @betatim).
  This is new in node 12.9, but the falsy behavior when this is undefined for old nodes should be fine, as it will just behave the same as it does now.
- More cases when we don't need to log a traceback for socket issues (EHOSTUNREACH, ETIMEDOUT)

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -425,15 +425,20 @@ class ConfigurableProxy extends EventEmitter {
     this.statsd.increment("requests." + code, 1);
     if (e) {
       // avoid stack traces on known not-our-problem errors:
-      // ECONNREFUSED (backend isn't there)
-      // ECONNRESET (backend is there, but didn't respond)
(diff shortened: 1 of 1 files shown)


meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/improved-chp-health-endpoint/9315/2

PR #291, opened 2021-03-05
adopt pre-commit

minrk:
- replaces `npm run fmt` which ran prettier on js
- autoformat html, markdown, yaml, scripts, not just js
- adds `eslint` and address a couple linter messages

The only reason *not* to do this is because pre-commit is a Python package and the js-community generally uses [husky](https://typicode.github.io/husky) to do things like pre-commit hooks.

On the other hand, the rest of JupyterHub uses pre-commit, and that's who contributes here.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,15 @@
+module.exports = {
+  env: {
+    commonjs: true,
+    es6: true,
+    node: true,
+  },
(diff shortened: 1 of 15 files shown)


consideRatio (approved):
LGTM, happy to see this spread across the JH org!

PR #293, opened 2021-03-05
Add tests for last-activity updates

minrk:
tests for #292 - both that the usual activity is unchanged for simple successful requests, and that activity is *not* updated for failed requests


consideRatio:
Node 10 LTS will reach end of life 2021-04-30, perhaps our next release should just stop supporting Node 10?

minrk:
I think I'd rather use Ubuntu LTS as baseline supported node, given how little we need from it, unless there's a big reason to cut people off. 16.04 reaches end-of-maintenance this April, at which point 18.04 will be the oldest LTS release, which ships with node 8.

PR #298, opened 2021-03-05
Simplify RELEASE.md with npm version

### LINKED ISSUE #276: Update release instructions
We have release instructions based on travis ci system and have not migrated documentation to the changes in #275

minrk:
Using npm version is a little simpler and more natural for a javascript project than bump2version.

closes #276


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "4.3.2-dev",
+  "version": "4.2.4-dev",
```
consideRatio, line 2:
```suggestion
  "version": "4.3.2-dev",
```

minrk: Yup, fixed. Ran through the actual steps just now to check that they work. Our bump2version wasn't getting package-lock.json which confused it.

RELEASE.md
```diff
@@ -13,69 +13,62 @@ meet the prerequisites:
...
 
    ```bash
-   # optionally first bump a minor or major version, but
-   # don't bump the patch version, that's already done.
-   # bump2version --no-commit minor
-   # bump2version --no-commit major
-   bump2version --tag release
+   # specify the new version here,
+   # which should already have been chosen when updating the changelog.
+   npm version 1.2.3
```
minrk, line 40: This in particular, explicitly stating the new version, means we can't publish the wrong thing, whereas `bump2version --tag release` assumes the *prior* state was correct for the next released version.

consideRatio (approved):
This looks great to me!

RELEASE.md
```diff
@@ -13,69 +13,62 @@ meet the prerequisites:
...
    ```
 
-1. Reset the version to the next development version with `bump2version`.
+1. Reset the version to the next development version with `npm version`,
+   without creating a tag:
 
    ```bash
-   bump2version --no-tag patch
+   npm version --no-git-tag-version prerelease --preid=dev
+   git commit -am "back to dev"
```
minrk, line 51: It still boggles my mind that `npm version` doesn't have a 'commit but don't tag' option

.npmrc
```diff
@@ -0,0 +1 @@
+tag-version-prefix=
```
minrk, line 1: `npm version`'s default is to use the `v` prefix on tags, but since we haven't been doing that (I don't happen like it, personally), this config lets us stay consistent.

RELEASE.md
consideRatio: I've come to appreciate that ;)

PR #308, opened 2021-04-24
Create Kj gf g

char1223:


consideRatio:
I asume this is cryptomining attempt...

PR #310, opened 2021-04-30
lib/mysql-store.js

mpermana:
This introduces MySQL backend storage, i.e:
```
bin/configurable-http-proxy  --storage-backend ./mysql-store --log-level debug
```

DDL:
```
CREATE TABLE routes (
    route  VARCHAR(300),
    target VARCHAR(300),
    PRIMARY KEY(route),
    INDEX index_target (target)
)
```

### DIFF
--- a/lib/mysql-store.js
+++ b/lib/mysql-store.js
@@ -1,0 +1,131 @@
+"use strict";
+
+var SqlString = require('sqlstring');
+const mysql = require('mysql');
+
+
+class MySQLStore {
+  constructor() {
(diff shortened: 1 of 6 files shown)


minrk:
Thanks for the PR! The redis backend is maintained as an [external package](https://github.com/globocom/configurable-http-proxy-redis-backend), and I'd like to follow that pattern for all storage backends, rather than bringing in unused storage implementations as dependencies in this package. What do you think about doing that?

mpermana:
sounds good to me, will close the PR

PR #316, opened 2021-05-26
Add changelog for 4.4.0

consideRatio:


manics:
Should this be a breaking change since it removes statsd support?

consideRatio:
Woops, hmmm, yes i guess. I have never heard of anyone using statsd but you are right.

I have published it as 4.4.0 now though, do we take an action or just go with it?

manics:
I don't know....

mriedem:
> 
> 
> Might have been good to mention that statsd metrics support was dropped in the same change. I didn't see a deprecation period for that or anything in the release notes.

Oops, saw this and commented in the commit, didn't notice this was already mentioned in the PR.

mriedem:
> I have published it as 4.4.0 now though, do we take an action or just go with it?

IMO at least having a minor version bump for the dropped support is minimum good. It would have been nice for anyone using those options to have a deprecation period before those options are dropped outright (we're not using them but might have, it was something I was going to investigate for monitoring the chp in our k8s cluster).

At least a reasonable follow up would be to amend the changelog to mention the dropped statsd support so people that are using it don't have to dig too far down into git / PR history to find what broke them.

PR #319, opened 2021-05-30
Set client SSL on only CA case

dtaniwaki:
I'd like to add routes to a server with self-signed TLS without client certification. I think it should be a reasonable use case with a networkpolicy in a Kubernetes cluster.


consideRatio:
I'm not able to think clearly about this change. I wonder if @manics has the understanding to confidently smash the merge button? This LGTM from what I can guess but I feel I don't really understand the implications of this change.

minrk (approved):

Issue #321, opened 2021-05-31
Failed to start Jupyter service with configurable-http-proxy 4.4.0


MvdB202:
### Bug description
With configurable-http-proxy version 4.4.0, JupyterHub failed to start proxy. Output:
``` 
/usr/lib/node_modules/configurable-http-proxy/node_modules/prom-client/lib/metrics/osMemoryHeapLinux.js:57
			} catch {
			        ^

SyntaxError: Unexpected token {
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/usr/lib/node_modules/configurable-http-proxy/node_modules/prom-client/lib/metrics/osMemoryHeap.js:4:22) 
```

#### Expected behaviour
I expected JupyterHub service to start as usual with the config proxy adding default route for Hub > http://127.0.0.1:8081

#### Actual behaviour
` May 31 10:37:46 my-instance jupyterhub[14230]: [I 2021-05-31 10:37:46.563 JupyterHub proxy:666] Starting proxy @ http://:8000 `

### How to reproduce
My instance is created with an ansible playbook where I had the npm module install configurable-http-proxy globally. Which now uses version 4.4.0 by default, which triggered this bug. 
_Currently I have version: '4.3.2' added to my playbook and everything is back up and running._

### Your personal set up
jupyterhub version 1.3.0
Python 3.6.9
Ubuntu 18.04
nginx 1.14.0

- <details><summary>Configuration</summary>

```python
# jupyterhub_config.py
c.Spawner.default_url = '/lab?reset'
c.Spawner.notebook_dir = '~'
c.Authenticator.admin_users = {'ubuntu'}
```
</details>

- <details><summary>Logs</summary>

```
# paste relevant logs here, if any
May 31 10:37:46 my-instance systemd[1]: Started Jypyter Hub daemon.
May 31 10:37:46 my-instance jupyterhub[14230]: [I 2021-05-31 10:37:46.507 JupyterHub app:2349] Running JupyterHub version 1.3.0
[...]

PR #323, opened 2021-06-04
Require node 12+ explicitly

### LINKED ISSUE #322: Require Node 12 (at least 10)
We specify that node >=6 is required, but in practice we test only against 10+ where tests are flaky on node 10.

Node 10 is no longer actively maintained, [only node 12+ is under active maintenance at the moment](https://nodejs.org/en/about/releases/). I suggest we bump our requirement from `node >= 6` to `node >= 12`.

@minrk and @manics what do you think? I believe our outdated specification of this relates to several issues we have run into in the past about TLS matters and now more recently perhaps also #321. Do you agree to require `node >= 12`?

https://github.com/jupyterhub/configurable-http-proxy/blob/790a8c7f91b26062e1184368a73560c20b782869/package.json#L26-L28

consideRatio:
We have failed to require a version of node which we test, which have resulted in failures for those with older versions of node. See for example #321 where node 8 caused problems. I suggest we transition properly to require node 12 at this point which is the oldest version of node with Long Term Support that is actively maintained still.

I figure we could cut a 5.0.0 release with this. In a way it is a bugfix to require node 10 as of 4.4.0 which broke with node 8, and it is a breaking change to require node 12 as node 10 have actually been functional I think. Note though that node 10 have had intermittent test failures while the other versions hasn't.

Since we have flaky tests for node 10, and I don't think our time is merited to debug and support that in detail any more, I suggest we go for node 12 and release CHP 5.0.0.

closes #322

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,8 +62,9 @@ jobs:
     strategy:
       fail-fast: false # Do not cancel all jobs if one fails
       matrix:
+        # IMPORTANT: Make sure to update package.json's enginges.node field to
+        #            always require at least the oldest version.
         node_version:
-          - "10"
           - "12"
           - "14"
           - "15"
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "ws": "^7.0.0"
   },
   "engines": {
-    "node": ">= 6.0"
+    "node": ">= 12.0"
   },
   "engineStrict": true,
   "main": "index.js",


manics (approved):
Sounds fine to me!

Are there any other potential breaking changes that should go in, or do we release 5.0.0 straight away?

consideRatio:
@manics not to my knowledge, I'm not feeling so capable pushing additional features or fixes if there were either though as I'm not feeling confident in writing code for this repo.

minrk:
👍 

Note that making a release dropping node support won't immediately or obviously affect any users with old node, since they will just get old CHP to match. We might also consider yanking 4.4 since it has an invalid dependency spec, and adding this restriction to 5.0 just means that users with node 8 will continue to get 4.4.

kgutwin:
Note that the README still says "Node >= 6" in the Install section.

consideRatio:
@kgutwin do you want to submit a PR fixing it? =)

PR #333, opened 2021-07-17
Fix to flags configuring ip addresses for ipv4+ipv6 compatebility

consideRatio:
Small changes to help z2jh avoid IPv4 and/or IPv6 issues by not specifying either `::` or `0.0.0.0` but instead a blank string when we want to listen far and wide as we do in a z2jh container.

### Review notes

I'd love to have this merged soon and cut a release of CHP to bump if for z2jh 1.1.0 that I'd like to make next week.

### Related
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2254
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2318

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -317,9 +317,9 @@ if (args.ip === "*") {
   args.ip = "";
 }
 listen.ip = args.ip;
-listen.apiIp = args.apiIp || "localhost";
+listen.apiIp = args.apiIp;
 listen.apiPort = args.apiPort || listen.port + 1;
-listen.metricsIp = args.metricsIp || "0.0.0.0";
+listen.metricsIp = args.metricsIp;
 listen.metricsPort = args.metricsPort;
 
 proxy.proxyServer.listen(listen.port, listen.ip);
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -87,7 +87,7 @@ cli
   )
   .option("--insecure", "Disable SSL cert verification")
   .option("--host-routing", "Use host routing (host as first level of path)")
-  .option("--metrics-ip <ip>", "IP for metrics server", "0.0.0.0")
+  .option("--metrics-ip <ip>", "IP for metrics server", "")
   .option("--metrics-port <n>", "Port of metrics server. Defaults to no metrics server")
   .option("--log-level <loglevel>", "Log level (debug, info, warn, error)", "info")
   .option(


consideRatio:
### Validation of logic script

```node
#!/usr/bin/env node

var cli = require("commander");
var log = require("../lib/log.js").defaultLogger();

var args = cli
  .option("--ip <ip-address>", "description...")
  .option("--api-ip <ip-address>", "description...", "localhost")
  .option("--metrics-ip <ip-address>", "description...", "")
  .parse(process.argv)
  .opts();

if (args.ip == "") {
    log.error("--ip was a blank string");
} else {
    log.error("--ip was: " + args.ip);
}

if (args.apiIp == "") {
    log.error("--api-ip was a blank string");
} else {
    log.error("--api-ip was: " + args.apiIp);
}

if (args.metricsIp == "") {
    log.error("--metrics-ip was a blank string");
} else {
    log.error("--metrics-ip was: " + args.metricsIp);
}
```

image

bin/configurable-http-proxy
```diff
@@ -317,9 +317,9 @@ if (args.ip === "*") {
   args.ip = "";
 }
 listen.ip = args.ip;
-listen.apiIp = args.apiIp || "localhost";
+listen.apiIp = args.apiIp;
 listen.apiPort = args.apiPort || listen.port + 1;
-listen.metricsIp = args.metricsIp || "0.0.0.0";
+listen.metricsIp = args.metricsIp;
```
consideRatio, line 322: Note that these were defaults that were set above during CLI parsing anyhow. By changing this, the actual change in logic outcomes are when we pass a blank string and `args.apiIp || "localhost"` no longer overrides the falsy blank string.

bin/configurable-http-proxy
```diff
@@ -87,7 +87,7 @@ cli
   )
   .option("--insecure", "Disable SSL cert verification")
   .option("--host-routing", "Use host routing (host as first level of path)")
-  .option("--metrics-ip <ip>", "IP for metrics server", "0.0.0.0")
+  .option("--metrics-ip <ip>", "IP for metrics server", "")
```
consideRatio, line 90: I've verified that apiIp became accessible when configured with a blank string just as if it would have been configured with `0.0.0.0`, but I think using a blank string, we also acquire the sensible IPv4 and/or IPv6 compatibility we want.

consideRatio:
### Validation of logic script

```node
#!/usr/bin/env node

var cli = require("commander");
var log = require("../lib/log.js").defaultLogger();

var args = cli
  .option("--ip <ip-address>", "description...")
  .option("--api-ip <ip-address>", "description...", "localhost")
  .option("--metrics-ip <ip-address>", "description...", "")
  .parse(process.argv)
  .opts();

if (args.ip == "") {
    log.error("--ip was a blank string");
} else {
    log.error("--ip was: " + args.ip);
}

if (args.apiIp == "") {
    log.error("--api-ip was a blank string");
} else {
    log.error("--api-ip was: " + args.apiIp);
}

if (args.metricsIp == "") {
    log.error("--metrics-ip was a blank string");
} else {
    log.error("--metrics-ip was: " + args.metricsIp);
}
```

image

PR #349, opened 2021-09-17
Restore node 10 as base supported engine

### LINKED ISSUE #348: npm engine requirement isn't taken into account at install time
In #323, we bumped the minimum node version, in part because I had [assumed](https://github.com/jupyterhub/configurable-http-proxy/pull/323#issuecomment-855644227) npm's node engine resolution worked like pip's (find the latest version we can run), but it doesn't. Instead, it [just warns that it won't work](https://www.stefanjudis.com/today-i-learned/prevent-npm-install-for-not-supported-node-js-versions/). It appears the best we can hope for is turning that warning into an error by adding strict engine checks, it cannot be considered an input to the installation request (which seems wild to me!).

Since the *latest* Ubuntu LTS (20.04) is still on node 10, I think we should bump minimum node back down to 10, which means dropping commander back to 7.x.

At the very least, in the future we should consider bumping the engine a major revision (which @consideRatio originally proposed, but I recommended against due to this misunderstanding), to make it easier to communicate the boundaries.

minrk:
reverts commander to 7.x

closes #348  which has more info: bumping the minimum engine is a high-impact change, and our needs are not high. Since our primary target is JupyterHub deployments and not node developers, our target baseline should be widely available without resorting to nodesource. To me, that means at the very least, the version in the most recent Ubuntu LTS should be supported, and that's node 10 in ubuntu 20.04 at the moment.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,11 +60,13 @@ jobs:
     strategy:
       fail-fast: false # Do not cancel all jobs if one fails
       matrix:
-        # IMPORTANT: Make sure to update package.json's enginges.node field to
+        # IMPORTANT: Make sure to update package.json's engines.node field to
         #            always require at least the oldest version, as well as our
         #            README.md file under the install section.
         node_version:
-          - "12" # Remove 2022-04-30, its end-of-life.
+          # Removing node 10 is dropping support for ubuntu 20.04 LTS
+          - "10"
+          - "12"
           - "14"
           - "15" # Remove about when 17 is available?
           - "16"
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ functionality to [JupyterHub] deployments.
 
 ## Install
 
-Prerequisite: [Node.js](https://nodejs.org/en/download/) ≥ 12
+Prerequisite: [Node.js](https://nodejs.org/en/download/) ≥ 10
 
 If you're installing `configurable-http-proxy` in Linux, you can follow [the instruction of nodesource](https://github.com/nodesource/distributions#installation-instructions) to install arbitrary version of Node.js.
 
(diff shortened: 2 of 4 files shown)


meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/installing-jupyterhub-on-ubuntu-20-04-server/10808/3

consideRatio:
Yikes, I'm a bit hesitant on this because I know of intermittent failures in our CI workflows showing up specifically for node 10, and that I worry that we invite for additional maintenance of a hard to debug problem.

I think we should support running jupyterhub on ubuntu 20.04, but out of the box without installing a node that isn't already end of life? Hmmm...

Have you also seen the intermittent failures in our CI system related to node 10? I didn't investigate them more in detail as I knew it was to be dropped before.

minrk:
node's "LTS" is not very "L" at all (2.5 years) which is barely the window between LTS releases. We don't require any recent node features, so I don't think we should be aggressive about dropping support for versions that work.

If node had better support for dropping old engine versions, I'd support tracking node LTS instead, but since all they have now is a completely unhelpful warning with no obvious action to take, and for our users, CHP is often the only node package they ever install, I think we should prioritize compatibility for users in realistic environments.

This is doubly frustrating now, because everything actually works fine with node 10, so the warning is purely a source of confusion and CHP works just fine on node 10, and the 'right' course of action for a user with node 10 is to just ignore the warning that CHP claims it is not compatible with node 10.

Flaky tests are frustrating, but if investigating the flakiness is not worth the necessary time, I think re-running flaky tests is preferable to dropping support for the latest widely available node version in our primary target distro.

consideRatio:
Thanks for providing some additional background for this decision! Let's go for it!

My worry isn't the test failure itself as much as I'm worried about always thinking that anyone reporting a bug could use node 10 and that the failure was a runtime failure that was related to the flakey test.

PR #369, opened 2022-02-02
fix custom-header

### LINKED ISSUE #368: `--custom-header` option doesn't work with only one header
Hi ! Thank you for jupyterhub 🙏 ❤️ !
I've been stumbling on a bug with this proxy. Probably introduced with #223 

### Bug description

The `--custom-header` flag doesn't work in version v4.5.0.
Running 
```shell
Proxy cmd: ['configurable-http-proxy', '--custom-header', 'testcustomheader:mytestcustomheader', '--log-level', 'debug', '--ip', '0.0.0.0', '--port', '8000', '--api-ip', '127.0.0.1', '--api-port', '8001', '--error-target', 'http://127.0.0.1:8081/hub/error']
```
Doesn't add the `testcustomheader:mytestcustomheader` header to proxified requests.

#### Expected behaviour
The headers contain the custom header set with the command-line option.

#### Actual behaviour
The headers don't contain the additional header. More over the parsed args are wrong:
```shell
Args: {
  apiIp: '127.0.0.1',
  xForward: true,
  prependPath: true,
  includePrefix: true,
  customHeader: undefined,
  metricsIp: '',
  logLevel: 'debug',
  ip: '0.0.0.0',
  port: 8000,
  apiPort: 8001,
  errorTarget: 'http://127.0.0.1:8081/hub/error'
}
```
notice the **undefined** for the `customHeader` option.

### How to reproduce
I've stumbled on this bug while using jupyterhub, so the setup is quite intricate and I don't have an easily reproducible setup at hand.
I can work on it though if needed.

### Your personal set up
I'm using `configurable-http-proxy` with jupyterhub with a non configurable reverse-proxy in front. So I can't setup the `X-Forwarded` as recommended by the docs. So I was trying to enforce them directly in `configurable-http-proxy` using the `--custom-header` flag.

 - OS: docker debian bullseye
 - Version(s): jupyterhub 2.0.2

- <details><summary>Configuration</summary>

```python
# jupyterhub_config.py
import sys
c.Application.log_level = 'DEBUG'
c.JupyterHub.authenticate_prometheus = False
c.JupyterHub.bind_url = 'http://0.0.0.0:8000'
c.JupyterHub.load_roles = [
    {
        "name": "jupyterhub-idle-culler-role",
        "scopes": [
            "list:users",
            "read:users:activity",
            "read:servers",
            "delete:servers",
            # "admin:users", # if using --cull-users
        ],
        # assignment of role's permissions to:
        "services": ["jupyterhub-idle-culler-service"],
    }
]
c.ConfigurableHTTPProxy.command = [
    'configurable-http-proxy',
    '--custom-header',
    'testcustomheader:mytestcustomheader',
    '--log-level', 'debug',
]
c.JupyterHub.services = [
    {
        "name": "jupyterhub-idle-culler-service",
        "command": [
            sys.executable,
            "-m", "jupyterhub_idle_culler",
            "--timeout=3600",
        ],
        # "admin": True,
    }
]
c.Spawner.debug = True
```
</details>

### Proposed fix

After some fidling, it appears the [collectHeadersIntoObject](https://github.com/jupyterhub/configurable-http-proxy/blob/main/bin/configurable-http-proxy#L110) is missing a return statement (and also calls an undefined log function in case of error).
Adding it fixes the issue.

```javascript
function collectHeadersIntoObject(value, previous) {
  var headerParts = value.split(":").map((p) => p.trim());
  if (headerParts.length != 2) {
    console.error("A single colon was expected in custom header: " + value);
    process.exit(1);
  }
  previous[headerParts[0]] = headerParts[1];
  return previous;
}
```

I'm willing to submit a PR if you are accepting them.

minrk:
collector must return the result

The test was passing because the update-in-place way we implemented the collector apparently works if you have _more than one_ header, but not if you have only one. I don't know if that's a bug upstream or not, but the right thing to do is to return after each collection.

Additional improvements:

- allow colons in header values (split only on first, since colons are quite sensible in header values, e.g. host:port)
- log and exit when proxy cli fails to start in tests

closes #368

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -111,7 +111,7 @@ function collectHeadersIntoObject(value, previous) {
   value = value.trim();
   var colon = value.indexOf(":");
   if (colon <= 0) {
-    log.error("A colon was expected in custom header: " + value);
+    console.error("A colon was expected in custom header: " + value);
     process.exit(1);
   }
   var headerName = value.slice(0, colon).trim();
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -108,12 +108,17 @@ cli
 // collects multiple flags to an object
 // --custom-header "k1:v1" --custom-header " k2 : v2 " --> {"k1":"v1","k2":"v2"}
 function collectHeadersIntoObject(value, previous) {
-  var headerParts = value.split(":").map((p) => p.trim());
-  if (headerParts.length != 2) {
-    log.error("A single colon was expected in custom header: " + value);
+  value = value.trim();
+  var colon = value.indexOf(":");
+  if (colon <= 0) {
+    log.error("A colon was expected in custom header: " + value);
     process.exit(1);
   }
-  previous[headerParts[0]] = headerParts[1];
+  var headerName = value.slice(0, colon).trim();
+  var headerValue = value.slice(colon + 1).trim();
+
+  previous[headerName] = headerValue;
+  return previous;
 }
 
 cli.parse(process.argv);
--- a/test/cli_spec.js
+++ b/test/cli_spec.js
@@ -16,6 +16,17 @@ function executeCLI(execCmd = "bin/configurable-http-proxy", args = []) {
   //cliProcess.stdout.pipe(process.stdout);
   const cliReady = new Promise((resolve, reject) => {
     var promiseResolved = false;
+    cliProcess.stderr.on("data", (data) => {
+      console.log(data.toString());
+    });
+    cliProcess.on("exit", (code) => {
+      if (!promiseResolved) {
+        console.log(
+          "process configurable-http-proxy " + args.join(" ") + "exited with code: " + code
+        );
+        reject();
+      }
+    });
     cliProcess.stdout.on("data", (data) => {
       if (data.includes("Route added /")) {
         defaultRouteAdded = true;
@@ -229,7 +240,7 @@ describe("CLI Tests", function () {
       "--custom-header",
       "k1: v1",
       "--custom-header",
-      " k2 : v2 v2 ",
+      " k2 : host:123 ",
     ];
     executeCLI(execCmd, args).then((cliProcess) => {
       childProcess = cliProcess;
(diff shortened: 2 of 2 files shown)


bin/configurable-http-proxy
```diff
@@ -108,12 +108,17 @@ cli
 // collects multiple flags to an object
 // --custom-header "k1:v1" --custom-header " k2 : v2 " --> {"k1":"v1","k2":"v2"}
 function collectHeadersIntoObject(value, previous) {
-  var headerParts = value.split(":").map((p) => p.trim());
-  if (headerParts.length != 2) {
-    log.error("A single colon was expected in custom header: " + value);
+  value = value.trim();
+  var colon = value.indexOf(":");
+  if (colon <= 0) {
+    log.error("A colon was expected in custom header: " + value);
```
Timost, line 114: I think calling `log` here isn't valid. I got crashes when fidling with this part of the code to make it log using `log`.

test/cli_spec.js
```diff
@@ -229,7 +240,7 @@ describe("CLI Tests", function () {
       "--custom-header",
       "k1: v1",
       "--custom-header",
-      " k2 : v2 v2 ",
+      " k2 : host:123 ",
```
Timost, line 243: Maybe adding a test with an invalid header (eg `--custom-header badheader`) would be worthwile ? I think the current error logging in `collectHeadersIntoObject` crashes.

Timost (commented):
Thank you for these changes ! 🙏 🙏 

Just a small doubt around the `log.error` call in `collectHeadersIntoObject`

bin/configurable-http-proxy
minrk: Good catch! Fixed and test added.

consideRatio (approved):
Wieee nice work @minrk and thank you @Timost for helping out with the review effort :tada: :rocket: :heart:!! This LGTM, great review points @Timost!

If this looks good to @Timost, let's merge!

Timost:
LGTM ! Thank you for doing all the work ! 🙏

PR #416, opened 2022-08-18
Updated alpine base image to 3.16

Christian-Nunnally:
The Alpine 3.15 image has security vulnerabilities according to the XRAY scanning tool. This PR updates the base image to 3.16 so that projects that depend on the configurable http proxy contain all the latest security updates.


consideRatio:
Thank you @Christian-Nunnally!

PR #430, opened 2022-09-08
Refine Dockerfile's labels and patch known vulnerabilities

consideRatio:


Dockerfile
```diff
@@ -1,10 +1,18 @@
 FROM node:lts-alpine3.16
 # ref: https://hub.docker.com/_/node?tab=tags&name=lts-alpine
 
-LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
```
consideRatio, line 4: The maintainer label was deprecated I recall

manics: I thought it was the `MAINTAINER` directive that was deprecated, not the label.

consideRatio: Yeah I recall that also the label was a deprecated practice. I've added the `org.opencontainers.image.authors` label now.

PR #451, opened 2022-12-05
Add changelog for 4.5.4

### LINKED ISSUE #450: Container rebuild to resolve open security issues
### Bug description

`jupyterhub/configurable-http-proxy:4.5.3 (alpine 3.16.2)` image includes critical (CVE-2022-42915) and high (CVE-2022-42916) vulnerabilities in `curl 7.83.1-r3`.

#### Expected behaviour

Fixable high or critical severity issues are resolved (automatically would be preferable)

#### Actual behaviour

A security scan with [trivy](https://trivy.dev) on the released version of `jupyterhub/configurable-http-proxy:4.5.3` returns the following issues:

```cmd
$ trivy image --ignore-unfixed jupyterhub/configurable-http-proxy:4.5.3
[…]
jupyterhub/configurable-http-proxy:4.5.3 (alpine 3.16.2)

Total: 6 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 2, CRITICAL: 2)

┌─────────┬────────────────┬──────────┬───────────────────┬───────────────┬────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Installed Version │ Fixed Version │                   Title                    │
├─────────┼────────────────┼──────────┼───────────────────┼───────────────┼────────────────────────────────────────────┤
│ curl    │ CVE-2022-42915 │ CRITICAL │ 7.83.1-r3         │ 7.83.1-r4     │ curl: HTTP proxy double-free               │
│         │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-42915 │
│         ├────────────────┼──────────┤                   │               ├────────────────────────────────────────────┤
│         │ CVE-2022-42916 │ HIGH     │                   │               │ curl: HSTS bypass via IDN                  │
│         │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-42916 │
│         ├────────────────┼──────────┤                   │               ├────────────────────────────────────────────┤
│         │ CVE-2022-32221 │ MEDIUM   │                   │               │ curl: POST following PUT confusion         │
│         │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-32221 │
├─────────┼────────────────┼──────────┤                   │               ├────────────────────────────────────────────┤
│ libcurl │ CVE-2022-42915 │ CRITICAL │                   │               │ curl: HTTP proxy double-free               │
│         │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-42915 │
│         ├────────────────┼──────────┤                   │               ├────────────────────────────────────────────┤
│         │ CVE-2022-42916 │ HIGH     │                   │               │ curl: HSTS bypass via IDN                  │
│         │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-42916 │
│         ├────────────────┼──────────┤                   │               ├────────────────────────────────────────────┤
│         │ CVE-2022-32221 │ MEDIUM   │                   │               │ curl: POST following PUT confusion         │
│         │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2022-32221 │
└─────────┴────────────────┴──────────┴───────────────────┴───────────────┴────────────────────────────────────────────┘
```

However, rebuilding the Dockerfile locally with no changes resolves this issue, as it [installs](https://github.com/jupyterhub/configurable-http-proxy/blob/3e1e0ec29a2d0f60191c44f531803aff33ddc789/Dockerfile#L15) the latest version of `curl`.

### How to reproduce

1. [Install](https://aquasecurity.github.io/trivy/v0.35/getting-started/installation/) `trivy`
2. Run `trivy image --ignore-unfixed jupyterhub/configurable-http-proxy:4.5.3`

### Requested fix

Build and publish `jupyterhub/configurable-http-proxy:4.5.4` using the current `main` branch.

Thanks!

consideRatio:
- By cutting this patch release, we can close #450

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,27 @@ command line for details.
 
 ## 4.5
 
+### [4.5.4] - 2022-12-06
+
+([full changelog](https://github.com/jupyterhub/configurable-http-proxy/compare/4.5.3...4.5.4))
+
+#### Dependency updates
+
+- build(deps): bump winston from 3.8.1 to 3.8.2 [#432](https://github.com/jupyterhub/configurable-http-proxy/pull/432) ([@dependabot](https://github.com/dependabot))
+
+#### Continuous integration
+
+- ci: small updates to workflows, updating misc versions [#439](https://github.com/jupyterhub/configurable-http-proxy/pull/439) ([@consideRatio](https://github.com/consideRatio))
+- build(deps): bump docker/build-push-action from 3.1.1 to 3.2.0 [#437](https://github.com/jupyterhub/configurable-http-proxy/pull/437) ([@dependabot](https://github.com/dependabot))
+- build(deps): bump docker/setup-buildx-action from 2.0.0 to 2.1.0 [#436](https://github.com/jupyterhub/configurable-http-proxy/pull/436) ([@dependabot](https://github.com/dependabot))
+- build(deps): bump docker/setup-qemu-action from 2.0.0 to 2.1.0 [#435](https://github.com/jupyterhub/configurable-http-proxy/pull/435) ([@dependabot](https://github.com/dependabot))
+
+#### Contributors to this release
+
+([GitHub contributors page for this release](https://github.com/jupyterhub/configurable-http-proxy/graphs/contributors?from=2022-09-09&to=2022-12-05&type=c))
+
+[@consideRatio](https://github.com/search?q=repo%3Ajupyterhub%2Fconfigurable-http-proxy+involves%3AconsideRatio+updated%3A2022-09-09..2022-12-05&type=Issues) | [@yuvipanda](https://github.com/search?q=repo%3Ajupyterhub%2Fconfigurable-http-proxy+involves%3Ayuvipanda+updated%3A2022-09-09..2022-12-05&type=Issues)
+
 ### [4.5.3] - 2022-09-09
 
 ([full changelog](https://github.com/jupyterhub/configurable-http-proxy/compare/4.5.2...4.5.3))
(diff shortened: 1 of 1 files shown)


minrk:
it's in a test dependency, so doesn't affect the release. The bumped package isn't in the image or any install of CHP.

I wish node made it easier to distinguish dev dependencies from real ones.

I don't think we should need to bump the package version _just_ to trigger image rebuilds, but the workflow may be the easiest way to do it. We should be able to rebuild/republish image tags, since the image tag is the version of the _package_, not the version of the _image_, and can be safely republished.

consideRatio:
Oh hmmm, what about having package-lock.json only include non dev dependencies if that is possible? Does that sound reasonable to consider further to you?

"npm install" doesnt respect the lock file, if i recall correctly, only "npm ci" does.

consideRatio:
To republish already tagged images is a controversial topic for us i recall. I'm undecided about it overall, but would rather for now keep adding a patch release to not need to reach agreement on a change in this case. I know its been discussed for z2jh images and the chp images in the past somewhere.

manics:
I'm also against overwriting the most specific tag, if we want to rebuild a release tag I'd prefer it if we added a `-N` suffix. 

We had a lot of discussion about the ideal way to build them in https://github.com/jupyterhub/team-compass/issues/474 but maybe we just go for something that works and refine the process later?

minrk:
I think it's reasonable to _expect_ security rebuilds of tagged images where the tag does not refer to the image, but something inside it. Docker image tags are not like git tags, they are more analogous to git branches and have reasonable expectations of updates unless the tag refers to the image itself (it does in our helm chart, it does not here or jupyterhub/jupyterhub or python:3.11.5, etc.).

I think the `-N` suffix is a good way to version the docker tag separately from the package. But I think what we have now is probably the best compromise of developer time vs sensible versions. The cost of publishing meaningless versions is pretty low (wasteful dependabot/conda-forge PRs, etc.) compared to the maintenance savings of having 1 package version == 1 docker image tag.

PR #491, opened 2023-08-17
Introduce keep alive and fix AWS Load Balancer 502 errors

a3626a:
I'm running Z2JH based service with about 1,000 DAU. It is deployed in AWS EKS attached to AWS ALB.
image

As DAU grows, users started to get 502 Responses from the LB.
image

This is well-known problem related to keep-alive setting. ([AWS Article](https://repost.aws/knowledge-center/elb-alb-troubleshoot-502-errors))
Unfortunately, configurable-http-proxy does not support keep-alive. So I implemented, and tested in production environment. 

After the deployment the number of 502 errors descreased.
image

Technical/Implementation detail

1) It is very important to allow keep-alive both client side and server side. That's why `Agent` and `keepAliveTimeout` are both needed.

2) The jupyter hub and jupyter server support keep-alive by default, because they are Tornado servers.

3) chp is given these parameters. They are AWS specific values.
```
"--server-keep-alive-timeout=61000"
"--agent-free-socket-timeout=62000"
```

PR #492, opened 2023-08-17
Enable keep-alive, `--keep-alive-timeout` flag added (default: 5000ms)

minrk:
extends #481 to enable keepAlive not just on the server, but on proxied requests as well via an http agent, and exposing timeout on the command-line.

alternative to #491 using the standard library

closes #481 
closes #491

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -103,6 +103,11 @@ cli
   .option(
     "--storage-backend <storage-class>",
     "Define an external storage class. Defaults to in-MemoryStore."
+  )
+  .option(
+    "--keep-alive-timeout <timeout>",
+    "Set timeout (in milliseconds) for Keep-Alive connections",
+    parseInt
   );
 
 // collects multiple flags to an object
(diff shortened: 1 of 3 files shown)


consideRatio:
I don't understand the details well enough to review this.

Are there two connections to consider (or more, or something else)? The connection initiated to the proxy (A->proxy), and the one the proxy initiated to some destination (proxy->B)?

minrk:
> Are there two connections to consider (or more, or something else)? The connection initiated to the proxy (A->proxy), and the one the proxy initiated to some destination (proxy->B)?

Yes, both of these. The `agent` is the pool used to allow keeping connections alive for the proxy->backend requests. The keepAlive argument is for requests _to_ the proxy. We need both of these because the keep-alive propagates all the way through the proxy, and if _either_ server (the proxy or the backend) sets `Connection: close` the connections will be closed all the way up. This is because the proxy server relays the `Connection` header to the lowest value.

If _any_ of the 4 entities involved (the client, the proxy's http server, the proxy's http client, or the upstream server) don't ask for the socket to be kept alive, then the connection will be closed.

- the client's response from the proxy will have Connection: close if _any_ of the following are true:
  - the client doesn't request keep-alive
  - the proxy server doesn't specify keepAlive support
  - the upstream response sets Connection: close
- and that upstream response has again the same logic, where the client must request (and the upstream server must support it, which tornado does).

consideRatio (approved):
Thanks a lot for helping me understand this better, I think I understand it well enough to dare agree this makes sense and merge given that this LGTM!

PR #508, opened 2023-10-18
Move from DockerHub to quay.io

yuvipanda:
Companion PR to
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3254. See that PR for more information.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -82,8 +82,8 @@ jobs:
         uses: jupyterhub/action-major-minor-tag-calculator@v2
         with:
           githubToken: ${{ secrets.GITHUB_TOKEN }}
-          prefix: "jupyterhub/configurable-http-proxy:"
-          defaultTag: jupyterhub/configurable-http-proxy:noref
+          prefix: "quay.io/jupyterhub/configurable-http-proxy:"
+          defaultTag: quay.io/jupyterhub/configurable-http-proxy:noref
 
       - name: Display tags
         run: echo "Docker tags ${{ steps.gettags.outputs.tags }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         #    https://github.com/jupyterhub/configurable-http-proxy/settings/secrets/actions
         if: startsWith(github.ref, 'refs/tags/')
         run: |
-          docker login -u "${{ secrets.DOCKERHUB_USERNAME }}" -p "${{ secrets.DOCKERHUB_TOKEN }}"
+          docker login -u "${{ secrets.QUAY_USERNAME }}" -p "${{ secrets.QUAY_PASSWORD }}" quay.io
```
minrk, line 71: This changes the creds, but doesn't change where it's pushed (next step in gettags needs `quay.io` added, I think)

yuvipanda: @minrk how about now?

minrk: I think that's right, thanks.

minrk (approved):

yuvipanda:
Thanks @minrk. I'll use workflow dispatch to run this just in case.

yuvipanda:
I just re-pushed 4.6.0 but realized that won't actually trigger the new workflow, only the old one.

I don't want to make a fresh release, so I've just re-run `➜ skopeo copy --multi-arch all docker://jupyterhub/configurable-http-proxy:4.6.0 docker://quay.io/jupyterhub/configurable-http-proxy:4.6.0` and am gonna call it done.

PR #516, opened 2023-11-27
Add changelog for 6.4.1

consideRatio:
I'm motivated to make this release for:
- verify that publishing to both dockerhub + quay works as intented
- have a fresh build with less known vulnerabilities and possibly now with node 20 instead of node 18 (we don't have that pinned)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,44 @@ command line for details.
 
 ## [Unreleased]
 
+### 4.6.1 - 2023-11-27
+
+([full changelog](https://github.com/jupyterhub/configurable-http-proxy/compare/4.6.0...4.6.1))
+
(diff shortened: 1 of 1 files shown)


minrk (approved):

consideRatio:
I got the PR title / commit title here wrong, it should be 4.6.1 and not 6.4.1 --- I'm publishing this as 4.6.1 now though!

consideRatio:
There was an issue with read only credentials to push to the image on quay.io that was fixed, but I never got the workflow to re-run correctly for some reason. It successfully logged in but still got 401. Doing it from my fork worked out though, so I figure it was perhaps re-using the read-only permissions or similar --- I'm confused...

I think this should work for jupyterhub/jupyterhub as well, and this repo going onwards --- but I'm not 100% sure.

PR #559, opened 2024-10-01
Replace `request` and `request-promise-native`

mjpsyapse:
`request` and `request-promise-native` have been deprecated for some time now. Removing them also resolves some CVE's even though they are dev dependencies.


minrk:
Thanks!

PR #574, opened 2025-05-15
migrate commonjs to esm

minrk:
This is mostly a find/replace of `require` to `import`, with little practical changes.

This is a breaking change for anyone who uses configproxy from another js package via undocumented `require('configurable-http-proxy')` API, but I think that is probably zero people, and now _technically_ only breaks on users of node 18, as you can `require` esm modules [starting with node 20](https://socket.dev/blog/require-esm-backported-to-node-js-20), which is the oldest supported LTS version by node.js itself, as of this month.

Trades coverage tool nyc for equivalent c8, since nyc doesn't support esm.

closes #563 by upgrading node-fetch, which is what prompted this change since it's an esm.

We could improve the code a lot by moving to async/await from chaining promises. That's purely internal, though, whereas this is an API change (though an undocumented, unsupported API that I _think_ nobody uses), so I wanted to do it as part of the already major bump due to the nodejs increase with http-proxy-3.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
         additional_dependencies:
           - setuptools
 
-  # Autoformat: markdown, javacsript
+  # Autoformat: markdown, javascript
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v4.0.0-alpha.8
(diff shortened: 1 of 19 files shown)


manics (approved):
Looks fine to me

PR #592, opened 2025-07-28
Support proxying to unix sockets

dometto:
## Summary

This PR adds support for proxying to servers that are listening on a Unix socket. The Node http client library already supports unix sockets, so adding this functionality was just a matter of setting the right values for the connection options.

It also adds a test for the new functionality.

## Rationale

[JupyterHub already supports listening on unix sockets](https://github.com/jupyterhub/jupyterhub/issues/3529#issuecomment-3073078980).

However, at this point, servers spawned by the hub have no way of contacting the Hub, as `configurable-http-proxy` is not capable of proxying back to the Hub when it listens on a unix socket. This functionality would therefore mean a significant step towards being able to host all JupyterHub services on a unix socket. Apart from some (probably small) performance benefits compared to TCP, this would also mean security gains in situations where endusers have shell access to the machine running Jupyter services (while servers listening on TCP can be contacted by any user on the system, unix sockets can be secured using normal file permissions).

A next step would be to allow `configurable-http-proxy` to listen on unix sockets itself, for both the proxy server and the API server. This should be pretty trivial, as Node's http server library already supports listening on unix sockets. This feature is already used in the test in this PR.

## Details

* The current implementation will treat any url with a protocol starting with `unix` as a unix socket path, so `unix+http://` (which is used by JupyterHub) and `unix://` will both work.
* When adding a route, it is important to pass in a url-encoded version of the path to the Unix socket: `unix+http://%2Ftmp%2F/test.sock`, and *not* `unix+http:///tmp/test.sock`. This again follows the current expectations of JupyterHub. When passing in the unencoded path, `URL.parse` will interpret `/tmp/test.sock` as the *path*, and set hostname to an empty string.


lib/configproxy.js
```diff
@@ -388,6 +388,34 @@ export class ConfigurableProxy extends EventEmitter {
     });
   }
 
+  proxyOptsForTarget(target, reqUrl) {
+    var proxyOptions = { target };
+
+    if (target.protocol.startsWith("unix")) {
+      proxyOptions.secure = false;
```
dometto, line 395: No support for SSL on unix sockets; this is not really needed anyway.

dometto:
@manics sorry for the direct ping (just looked at the last active contributor). Could you let me know if there is any interest in this feature?

Issue #557, opened 2024-09-19
Hundred of users leads to running out of tens of thousands of ephemeral ports


consideRatio:
From https://github.com/jupyterhub/configurable-http-proxy/issues/388#issuecomment-2359217928 and onwards is context on how a CHP pod can end up running out of ephemeral ports, with a mitigation strategy in https://github.com/jupyterhub/configurable-http-proxy/issues/388#issuecomment-2362097477.

consideRatio:
This is from a CHP pod with a hub currently having 222 current user pods running the image `quay.io/2i2c/utoronto-image:2525722ac1d5`, where users may be accessing /tree or /lab and its not clear what distribution of UI usage among those.

```
/srv/configurable-http-proxy $ netstat -natp | grep ESTABLISHED | grep 8081 | wc -l
80
/srv/configurable-http-proxy $ netstat -natp | grep ESTABLISHED | grep 8888 | wc -l
1416
/srv/configurable-http-proxy $ netstat -natp | grep ESTABLISHED | wc -l
1609
```

From inspection, it seems this makes use of jupyter_server 2.12.1 and jupyterlab 4.0.9.

<details><summary><code>pip list</code></summary>

```
Package                           Version
--------------------------------- ------------
absl-py                           2.1.0
affine                            2.4.0
aiohttp                           3.9.5
aiosignal                         1.3.1
alabaster                         0.7.16
alembic                           1.13.0
altair                            5.2.0
annotated-types                   0.7.0
anyio                             4.1.0
archspec                          0.2.2
argon2-cffi                       23.1.0
argon2-cffi-bindings              21.2.0
arrow                             1.3.0
arviz                             0.18.0
astropy                           5.3.4
astroquery                        0.4.7
asttokens                         2.4.1
astunparse                        1.6.3
async-generator                   1.10
async-lru                         2.0.4
async-timeout                     4.0.3
attrs                             23.1.0
Babel                             2.13.1
backports.tarfile                 1.0.0
beautifulsoup4                    4.12.2
bleach                            6.1.0
blinker                           1.7.0
blis                              0.7.10
bokeh                             3.3.2
boltons                           23.0.0
Bottleneck                        1.3.7
branca                            0.7.2
Brotli                            1.1.0
cached-property                   1.5.2
cachetools                        5.4.0
catalogue                         2.0.10
certifi                           2023.11.17
certipy                           0.1.3
cffi                              1.16.0
charset-normalizer                3.3.2
[...]

PR #598, opened 2025-09-19
close proxied websockets when upstream sends a regular response

minrk:
this is a band-aid for #557 

the 'true' fix ought to be https://github.com/sagemathinc/http-proxy-3/pull/27

I don't think we should merge this, I'm just showing folks what it looks like, in case anyone wants to test this before the release is available


consideRatio:
@minrk is this PR still relevant or not?

PR #646, opened 2026-05-08
feat: Allow --error-path fallback when --error-target error fetch fails

kaddynator:
**Problem:**

Today you must choose either --error-target or --error-path. If --error-target is set and the request to the custom error server fails (hub down, connection refused, etc.), CHP only returns the generic plain-text response. Static HTML under --error-path is never used in that case.

**Change:**

- Permit both --error-target and --error-path. When --error-target is configured, CHP still tries the custom error URL first. On request error, serve HTML from --error-path using the existing rules ({code}.html, then error.html, else default).

-  Fix --error-path CLI help: it is a directory of HTML files, not another proto://host server.

Add a test that uses an unreachable errorTarget and asserts responses come from the error-path fixtures.

**Backward compatibility:**
Using only --error-target or only --error-path behaves as before. If both are set, behavior is: dynamic page when reachable; otherwise file-based fallback.

### DIFF
--- a/bin/configurable-http-proxy
+++ b/bin/configurable-http-proxy
@@ -70,7 +70,10 @@ cli
     "--error-target <host>",
     "Alternate server for handling proxy errors (proto://host[:port])"
   )
-  .option("--error-path <path>", "Alternate server for handling proxy errors (proto://host[:port])")
+  .option(
+    "--error-path <path>",
+    "Directory containing {status}.html (and optional error.html) for proxy errors"
+  )
   .option(
     "--redirect-port <redirect-port>",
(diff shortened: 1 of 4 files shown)


minrk:
This seems like a reasonable goal!

Sorry we haven't communicated it widely yet, so it wasn't clear already before the PR was opened, but we have adopted an [LLM policy](https://compass.hub.jupyter.org/contribute/llm/) that requires human authorship of all code and documentation and communication. Since this PR has commits attributed to Cursor, it violates that policy and cannot be accepted, so I will close it. If you'd like to contribute to JupyterHub in the future, please feel free to read our [contributing documentation](https://compass.hub.jupyter.org/contribute/guide/).

PR #648, opened 2026-05-21
feat: timeout error-target fetch and fall back to error-path

kaddynator:
# Problem:

Today you must choose either --error-target or --error-path. If --error-target is set and the request to the custom error server fails (hub down, connection refused, etc.), CHP only returns the generic plain-text response. Static HTML under --error-path is never used in that case.

# Change:

Permit both --error-target and --error-path. When --error-target is configured, CHP still tries the custom error URL first. On request error, serve HTML from --error-path using the existing rules ({code}.html, then error.html, else default).

Fix --error-path CLI help: it is a directory of HTML files, not another proto://host server.

Add a test that uses an unreachable errorTarget and asserts responses come from the error-path fixtures.

# Backward compatibility:
Using only --error-target or only --error-path behaves as before. If both are set, behavior is: dynamic page when reachable; otherwise file-based fallback.

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -196,6 +196,16 @@ export class ConfigurableProxy extends EventEmitter {
     }
     this.errorPath = options.errorPath || path.join(__dirname, "error");
 
+    var eto =
+      options.errorTargetTimeout ?? options.error_target_timeout ?? this.options.errorTargetTimeout;
+    if (eto === undefined || eto === null) {
+      eto = 10000;
+    }
+    this.errorTargetTimeout = typeof eto === "string" ? parseInt(eto, 10) : Number(eto);
+    if (!Number.isFinite(this.errorTargetTimeout) || this.errorTargetTimeout < 1) {
+      this.errorTargetTimeout = 10000;
+    }
+
     if (this.options.enableMetrics) {
       this.metrics = new metrics.Metrics();
     } else {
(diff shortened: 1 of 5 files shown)


lib/configproxy.js
```diff
@@ -196,6 +196,16 @@ export class ConfigurableProxy extends EventEmitter {
     }
     this.errorPath = options.errorPath || path.join(__dirname, "error");
 
+    var eto =
+      options.errorTargetTimeout ?? options.error_target_timeout ?? this.options.errorTargetTimeout;
+    if (eto === undefined || eto === null) {
+      eto = 10000;
+    }
+    this.errorTargetTimeout = typeof eto === "string" ? parseInt(eto, 10) : Number(eto);
```
minrk: at this point, timeout should be an int, no need to handle the string here

lib/configproxy.js
```diff
@@ -196,6 +196,16 @@ export class ConfigurableProxy extends EventEmitter {
...
+    if (!Number.isFinite(this.errorTargetTimeout) || this.errorTargetTimeout < 1) {
```
minrk: Let's remove this check. Rather than ignoring invalid values, we should assign what is given. If the value is invalid and leads to an error, an error is the right result.

lib/configproxy.js
VishnuVV27: Addressed in 2fc0b9a. Thanks for the review!

lib/configproxy.js
VishnuVV27: Addressed in 2fc0b9a. Thanks for the review!

PR #654, opened 2026-06-06
feat(store): Add Redis-backed route store to persist routes across restarts

osmangoninahid:
The routing table previously lived only in process memory, so a restart (e.g. a Kubernetes pod reschedule) lost all dynamically-registered routes.

Add an opt-in Redis store, selected with '--store redis' (REDIS_URL env var required; default remains the in-memory store). RedisStore extends MemoryStore: reads stay in-memory via the URLTrie so the proxy hot path is unchanged, while add/update/remove are mirrored to a Redis hash and the table is hydrated from Redis on startup. last_activity is revived as a Date on load.

lib/store.js: RedisStore (node-redis)
lib/configproxy.js: loadStorage selects RedisStore on options.store === 'redis'
bin: --store flag + REDIS_URL validation
tests: redis_store_spec.js (incl. restart-persistence), proxy_spec wiring case
CI: redis service container in test job
README: --store option + 'Persisting routes with Redis' section

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,19 @@ jobs:
 
   test:
     runs-on: ubuntu-24.04
+    # Redis service so the RedisStore tests run; REDIS_URL is set below.
+    services:
+      redis:
+        image: redis:7
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    env:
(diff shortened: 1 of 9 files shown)


minrk:
Thanks for the PR!

We've had this feature for some time, since we had pluggable backend storage with https://github.com/globocom/configurable-http-proxy-redis-backend

but I'm not sure anyone uses it, since it hasn't been updated in a long time.

I've you'd like to update the redis support, you can look at that bringing that package up-to-date.

PR #655, opened 2026-06-17
Standardize unix socket protocol to http+unix

dometto:
As discussed in https://github.com/jupyterhub/jupyterhub/pull/5397, it would be good to standardize the protocol used for unix sockets across jupyter projects. `jupyter-server` already uses `http+unix`, as do other popular projects.

This PR replaces unix+http with http+unix throughout this project.

I think the only consequence is that currently, the implementation technically supports `unix+https` (https on unix socket), since we check for URLs *starting* with `"unix+"`. However, checking for URLs that start with `"http+unix"` excludes `"https+unix"`. We could check for `http+unix` *or* `https+unix`, but I wonder if this is worth it. SSL over unix sockets is technically possible, I think, but I don't know why anyone would really want to use it...

### DIFF
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -538,7 +538,7 @@ export class ConfigurableProxy extends EventEmitter {
       var url = "";
       if (options.target.socketPath) {
         options.target.hostname = "localhost";
-        url = options.target.toString().replace("http+unix", "http")
+        url = options.target.toString().replace("http+unix", "http");
       } else {
         url = options.target.toString();
       }
--- a/lib/configproxy.js
+++ b/lib/configproxy.js
@@ -424,7 +424,7 @@ export class ConfigurableProxy extends EventEmitter {
   proxyOptsForTarget(target) {
     var proxyOptions = { target };
 
-    if (target.protocol === "http+unix" || target.protocol === "unix+http") {
+    if (target.protocol.startsWith("http+unix") || target.protocol.startsWith("unix+http")) {
       // No need for agents for unix sockets
       // No support for https for unix sockets
       proxyOptions.secure = false;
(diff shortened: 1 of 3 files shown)


dometto:
Test failures don't look specific to this PR.

lib/configproxy.js
```diff
@@ -424,7 +424,7 @@ export class ConfigurableProxy extends EventEmitter {
   proxyOptsForTarget(target) {
     var proxyOptions = { target };
 
-    if (target.protocol.startsWith("unix")) {
+    if (target.protocol.startsWith("http+unix")) {
```
minrk: I think we can use equality to check this.

For backward-compatibility, maybe we should accept `unix+http` to avoid needing a major release:


```suggestion
    if (target.protocol === "http+unix" || target.protocol === "unix+http") {
```

lib/configproxy.js
```diff
@@ -538,7 +538,7 @@ export class ConfigurableProxy extends EventEmitter {
       var url = "";
       if (options.target.socketPath) {
         options.target.hostname = "localhost";
-        url = options.target.toString().substring(5); // chop off unix+
+        url = options.target.toString().replace("http+unix", "http");
```
minrk: for backward-compatibility:

```suggestion
        url = options.target.toString().replace("http+unix", "http").replace("unix+http", "http");
```

minrk (commented):
This is great! The only thing I think is that since we released `unix+http`, we should still accept that. We can deprecate it, but this avoids an unnecessary breakage.

I suggested two fixes required to resolve it, but we could also normalize to `http+unix` when we receive target URLs, so that we guarantee everywhere internally that it is consistent.

minrk:
The CI failures have been fixed in main

lib/configproxy.js
dometto: Applied this suggestion but tests were failing, also looking at the `else if` clauses below this line it appears we have to use `startsWith` after all (`protocol` includes the `://` part I think).

Reverted to `startsWith`, but with the backward compatibility clause in the if-statement.

dometto:
Thanks for the review! Tests are passing now.

PR #659, opened 2026-07-08
changelog for 5.3

minrk:
release with #655 so we can proceed with https://github.com/jupyterhub/jupyterhub/pull/5397


consideRatio (approved):
❤️

PR #670, opened 2026-09-15
require node 22

minrk:
### Checklist

- [x] I am the author of this work

### What does this PR do?

Type of change:
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [x] Maintenance (testing, packaging, infrastructure, metadata)
- [ ] Other

drop node 20, require node 22. node 20 reached end of life a few months ago.

### Is this PR related to an issue, or is it part of a larger body of work?

bumping minimum node to 22 is required for #668 

### Does this PR introduce a breaking change?

No, updating the minimum node is not a breaking change, but it does affect which environments will continue to get future updates.

### What should a reviewer concentrate their feedback on?

Maybe check if there's any lingering references to node 20 that I missed, like #666


manics (approved):
Should we treat this as a breaking change?

minrk:
I don't think so. node handles it fine, just like Python, so folks on unsupported node will get the latest version that does support their runtime. I'd call it a minor bump.

